PR #10, opened 2025-01-14
Add module for microgrid configs

cwasicki:
Includes microgrid config that contains component configs to get component IDs and formulas for "asset-level" component types like pv, battery, load, etc. and metadata information for microgrids like gridpool ID or delivery area.


cyiallou (approved):
looks good to me!

PR #16, opened 2025-01-21
Support component_category in type IDs retrieval

cyiallou:
Add `component_category` input argument to `component_type_ids` method of `MicrogridConfig`.
- Define valid categories: "meter", "inverter", "component".
- Default to `cfg.cids()` when no category is specified.
- Handle cases where the specified category is invalid or has no IDs.

This change adds flexibility in retrieving IDs for specific component categories while maintaining existing behaviour.

### DIFF
--- a/src/frequenz/lib/notebooks/config.py
+++ b/src/frequenz/lib/notebooks/config.py
@@ -10,6 +10,9 @@ from typing import Any, Literal, cast, get_args
 ComponentType = Literal["grid", "pv", "battery", "load", "chp"]
 """Valid component types."""
 
+ComponentCategory = Literal["meter", "inverter", "component"]
+"""Valid component categories."""
+
 
 @dataclass
 class ComponentTypeConfig:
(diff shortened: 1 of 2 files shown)


src/frequenz/lib/notebooks/config.py
```diff
@@ -126,22 +129,45 @@ def component_types(self) -> list[str]:
...
 
+        if component_category:
+            valid_categories = get_args(ComponentCategory)
+            if component_category not in valid_categories:
+                raise KeyError(
+                    f"Invalid component category: {component_category}. "
+                    f"Valid categories are {valid_categories}"
+                )
+            category_ids = cast(
+                list[int] | None, getattr(cfg, component_category, None)
```
cwasicki: Can't the default be an empty list?

src/frequenz/lib/notebooks/config.py
```diff
@@ -126,22 +129,45 @@ def component_types(self) -> list[str]:
...
+            )
+            if category_ids:
+                return category_ids
+            raise LookupError(
```
cwasicki: I am not sure if we should raise here. E.g. having a microgrid without an inverter is totally fine, whether this is a problem or not depends on the use-case.

cyiallou: You’re right that having a microgrid without certain component categories can be valid depending on the use case. My initial reasoning for raising an exception here was to ensure that missing IDs for a specified category would be flagged explicitly, helping to catch potential configuration issues early and avoid silent failures. However, I understand that this strict validation may not align with all use cases. Returning an empty list instead would provide more flexibility, allowing downstream code to handle the absence of components as needed without assuming it's an error.

cyiallou:
updated

cwasicki (approved):

PR #17, opened 2025-01-21
Add assets config to microgrid config

cwasicki:
Config files specifying assets pv, wind and battery are now supported and asset metadata can be accessed via the `assets` property. For each asset type multiple entries can be specified in dict form.

### DIFF
--- a/src/frequenz/lib/notebooks/config.py
+++ b/src/frequenz/lib/notebooks/config.py
@@ -74,6 +74,53 @@ class ComponentTypeConfig:
 
 
 @dataclass(frozen=True)
+class PVConfig:
+    """Configuration of a PV system in a microgrid."""
+
+    peak_power: float | None = None
+    """Peak power of the PV system in Watt."""
+
+    rated_power: float | None = None
+    """Rated power of the inverters in Watt."""
+
+
+@dataclass(frozen=True)
+class WindConfig:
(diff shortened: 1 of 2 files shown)


src/frequenz/lib/notebooks/config.py
```diff
@@ -111,6 +161,10 @@ def __init__(self, config_dict: dict[str, Any]) -> None:
         """
         self._metadata = Metadata(**(config_dict.get("meta") or {}))
 
+        self._assets_cfg = AssetsConfig(
+            pv=config_dict.get("pv") or {},
+        )
```
cyiallou: Is there a specific reason for only including `pv` for now?

I suggest updating the `__init__` method to initialize `_assets_cfg` with configurations for all three asset types. This would make the behaviour consistent with the `AssetsConfig` design.

cyiallou (commented):
Nice work on adding support for asset configurations and integrating `_assets_cfg` into `MicrogridConfig` is a solid addition. I can see how this will make the configuration more modular and extendable in the future. I only have a minor comment/suggestion.

src/frequenz/lib/notebooks/config.py
cwasicki: Good point, will update

cwasicki:
Updated.

cyiallou (approved):

PR #27, opened 2025-02-19
Allow missing ctype field in microgrid config

cyiallou:
Updated `MicrogridConfig` to handle missing `ctype` fields, allowing collection of incomplete microgrid configs.


cwasicki (approved):
Thanks a lot for fixing :)

PR #28, opened 2025-02-24
Support metric dependent formulas

cwasicki:
This supports having defining different formulas for each metric in the config file. An example for different formulas would be power values, which should be summed, and SoC, where a sum doesn't make sense. Instead a capacity-weighted mean can be used for this.

Moreover, the metric name can be passed to the `cids` method to support logic depending on the metric name. Since the method is used to allow getting all CIDs that are required to calculate certain component type metrics, the CIDs can differ depending on the metric that is calculated.

Without passing a metric, the logic of the method is unchanged. If a metric was passed, the method extracts the CIDs from the formula. The formula can be either a user-defined formula or a default formula if available.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,12 +2,15 @@
 
 ## Summary
 
-Renamed the `"load"` component type to `"consumption"` for clarity.
 
 ## Upgrading
 
-- If your code references `"load"` as a component type, update it to `"consumption"`.
-- Made the `MicrogridConfig` reader tolerant to missing `ctype` fields, allowing collection of incomplete microgrid configs.
+* Made the `MicrogridConfig` reader tolerant to missing `ctype` fields, allowing collection of incomplete microgrid configs.
+* Formula configs are now defined per metric to support different formulas for each metric in the same config file.
+  This is a breaking change which requires updating the formula fields in the config file.
+* Default formulas are defined for AC active power and battery SoC metrics.
+  The default SoC calculation uses simple averages and ignore different battery capacities.
+* The `cids` method is changed to support getting metric-specific CIDs which in this case are extracted from the formula.
 
 ## New Features
 
--- a/src/frequenz/lib/notebooks/config.py
+++ b/src/frequenz/lib/notebooks/config.py
@@ -41,6 +41,12 @@ class ComponentTypeConfig:
             self.formula["AC_ACTIVE_POWER"] = "+".join(
                 [f"#{cid}" for cid in self._default_cids()]
             )
+        if self.component_type == "battery" and "BATTERY_SOC_PCT" not in self.formula:
+            if self.component:
+                cids = self.component
+                form = "+".join([f"#{cid}" for cid in cids])
+                form = f"({form})/({len(cids)})"
+                self.formula["BATTERY_SOC_PCT"] = form
 
     def cids(self, metric: str = "") -> list[int]:
         """Get component IDs for this component.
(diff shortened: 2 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -2,12 +2,11 @@
 
 ## Summary
 
-Renamed the `"load"` component type to `"consumption"` for clarity.
 
 ## Upgrading
 
-- If your code references `"load"` as a component type, update it to `"consumption"`.
-- Made the `MicrogridConfig` reader tolerant to missing `ctype` fields, allowing collection of incomplete microgrid configs.
+* Formula configs are now defined per metric to support different formulas for each metric in the same config file. A
```
flora-hofmann-frequenz: There's a rogue 'A' at the end of the sentence.

src/frequenz/lib/notebooks/config.py
```diff
@@ -30,20 +31,52 @@ class ComponentTypeConfig:
...
+            ValueError: If no IDs are available or no formula exists for the metric.
+        """
+        if metric:
+            assert isinstance(self.formula, dict)
+            formula = self.formula.get(metric)
+            if not formula:
+                raise ValueError(
+                    f"{metric} does not have a formula for {self.component_type}"
+                )
+            pattern = r"#(\d+)"
```
flora-hofmann-frequenz, line 77: maybe a little comment would help here to explain the pattern for an example #123 or something. Not sure if everyone is regex literate.

flora-hofmann-frequenz (dismissed):
Just two smaller comments that could be worked in, otherwise LGTM.

cyiallou (commented):
Besides the two comments from Flora, it looks good to me.

cwasicki:
@flora-hofmann-frequenz @cyiallou Updated according to your comments and fixed some issues to pass the new CI.

RELEASE_NOTES.md
```diff
@@ -2,12 +2,14 @@
 
 ## Summary
 
-Renamed the `"load"` component type to `"consumption"` for clarity.
 
 ## Upgrading
 
-- If your code references `"load"` as a component type, update it to `"consumption"`.
-- Made the `MicrogridConfig` reader tolerant to missing `ctype` fields, allowing collection of incomplete microgrid configs.
```
cyiallou, line 10: The bullet point "Made the `MicrogridConfig`" should not be deleted. There was a mistake in a previous PR (see commit: https://github.com/frequenz-floss/frequenz-lib-notebooks/commit/f6939c6e1d763dea6cfd9d347b3ebd8fb0e73019) where the first bullet point, which belongs to the previous release, got deleted and added back with the referenced commit.

src/frequenz/lib/notebooks/config.py
```diff
@@ -240,11 +235,13 @@ def formula(self, component_type: str, metric: str) -> str:
         cfg = self._component_types_cfg.get(component_type)
         if not cfg:
             raise ValueError(f"{component_type} not found in config.")
-
-        if not cfg.has_formula_for(metric):
+        if cfg.formula is None:
+            raise ValueError(f"No formula set for {component_type}")
```
cyiallou, line 281: Could we add all the different `ValueError`s to the docstring for completeness please? :)

cyiallou (changes requested):
The logic looks good to me. Just a minor comment on the release notes and some cosmetic changes.

cwasicki:
@cyiallou Fixed

cyiallou (approved):

PR #30, opened 2025-02-28
Open source basic Alert and Solar Notebooks

cyiallou:


cyiallou:
Instruction documents are still needed.

cyiallou:
Rebased to latest main branch

cyiallou:
Added the instruction docs.

cyiallou:
Replaced a hardcoded value with the corresponding variable (a remnant from testing)

cwasicki:
Couple of comments:
* The notebooks have execution errors. Better make sure that the committed versions are working out-of-the box and remove all cell outputs.
* I think we should only have a single notebook with both languages, maintaining two versions of the same notebook can easily go out of sync. If you feel it gets too bloated maybe we can move some text out of the notebook again? The original idea to have a lot of doc there originates from the time where we were planning to send self-contained notebooks via mail. What do you think?
* Is it possible to move more of the logic in the alerting notebook into the library?

cyiallou:
> The notebooks have execution errors. Better make sure that the committed versions are working out-of-the box and remove all cell outputs.

The notebooks do work out of the box. The execution errors you see are basically the Deepnote UI elements which display some variable values and for the deepnote plots.

> I think we should only have a single notebook with both languages, maintaining two versions of the same notebook can easily go out of sync. 

I agree.

> If you feel it gets too bloated maybe we can move some text out of the notebook again? The original idea to have a lot of doc there originates from the time where we were planning to send self-contained notebooks via mail. What do you think?

Personally, I prefer to have self-contained notebooks instead of having to read a separate document in order to understand how to execute it. On the other hand, I'd hate having to read a bunch of text too. 

> Is it possible to move more of the logic in the alerting notebook into the library?

I suppose so. For example, `generate_html_email` is a large function which I created so that the user can have an example template email (as a starting point for them to customise). It can be moved into the library. I'll check other parts too.


**Next steps:**
- I'll check if I can remove the deepnote cell outputs (the ones with execution error).
- Merge the notebooks into 1 for both languages.
- Move some of the logic out of the alerting notebook.

src/frequenz/lib/notebooks/alerts/alert_email.py
```diff
@@ -0,0 +1,304 @@
...
+            ),
+            unique_states=(
+                "state_value",
+                lambda x: [html.escape(str(s)) for s in x.unique()],
+            ),
+            unique_components=("component_id", lambda x: list(x.unique())),
+        )
+        .reset_index()
+    )
+
```
cwasicki, line 191: There could be an option to filter out groups without any error or warning.

cyiallou: Good idea. By default or as an option for the user?

cwasicki: Maybe as an option for the user but True by default. For some customer I could imagine that they would still be interested in seeing which components were analyzed.

cwasicki (dismissed):

cwasicki (approved):

PR #37, opened 2025-03-03
Fix flake8 issues

cyiallou:
The following changes were made to proactively resolve issues that will be raised in the new linter versions.

Fixes the issues in PR #35

### DIFF
--- a/src/frequenz/lib/notebooks/solar/maintenance/data_processing.py
+++ b/src/frequenz/lib/notebooks/solar/maintenance/data_processing.py
@@ -135,12 +135,13 @@ def preprocess_data(
     time_diff_seconds_series = time_diff_seconds_arr[
         time_diff_seconds_arr.notna() & (time_diff_seconds_arr != 0)
     ]
-    try:
-        assert (
-            time_diff_seconds_series.nunique() == 1
-        ), "Time differences are not constant. Taking the most frequent value."
+    if time_diff_seconds_series.nunique() == 1:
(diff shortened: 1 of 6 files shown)


src/frequenz/lib/notebooks/solar/maintenance/data_processing.py
```diff
@@ -135,12 +135,9 @@ def parse_unit(unit_str: str) -> tuple[float, str]:
...
     ]
-    try:
-        assert (
-            time_diff_seconds_series.nunique() == 1
-        ), "Time differences are not constant. Taking the most frequent value."
+    if time_diff_seconds_series.nunique() == 1:
         time_diff_seconds = time_diff_seconds_series.values[0]
-    except AssertionError:
+    else:
         time_diff_seconds = time_diff_seconds_series.mode()[0]
```
ghost, line 145: Is it expected that `time_diff_seconds_series` can have multiple unique values, or is this an anomaly? Since it would previously raise an assertion error, should we add a warning if this is an unexpected behaviour?

cyiallou: You are right, multiple unique values in `time_diff_seconds_series` should probably not happen and could indicate some kind of an anomaly in the data. I suspect this case was introduced when the function needed to handle both reporting API data and test files from past development. To address this, I've added a warning so that users are alerted when it occurs.

ghost (approved):
LGTM!

PR #40, opened 2025-03-20
Avoid importing CSSDict from Pandas to prevent Jinja2 conflicts

cyiallou:
Replaced the import of `CSSDict` with `cast(Any, ...)` to avoid dependency conflicts with `Jinja2` in `Deepnote`. This maintains compatibility with `pandas.Styler.set_table_styles()` while removing unnecessary imports. No functional changes.


flora-hofmann-frequenz (approved):
LGTM

PR #42, opened 2025-03-24
Add simple PV forecast using weather API

cwasicki:
This notebook shows how to use the frequenz.client.weather package to get solar irradiance forecasts and use it to create a simple PV forecast.


cyiallou (approved):
Looks good to me. Useful example.

PR #61, opened 2025-04-30
Add module for microgrid component data 

cwasicki:
Introduces the MicrogridData class which brings functionality to fetch ac active power and SoC data aggregate over multiple components of the same type from the reporting API.

Also moves microgrid config to new data namespace.


src/frequenz/data/microgrid/component_data.py
```diff
@@ -0,0 +1,233 @@
...
+from frequenz.client.common.metric import Metric
+from frequenz.client.reporting import ReportingApiClient
+
+from .config import MicrogridConfig
+
+_logger = logging.getLogger(__name__)
+
+
+class MicrogridData:
+    """Fetch power data of for component types of a microgrid."""
```
cyiallou: "of" is a typo here

RELEASE_NOTES.md
cyiallou, line 1: The release notes need to be cleared before you add the new notes.

cyiallou (changes requested):
I am confused with the second commit message. In that commit you introduce the class `MicrogridData` but the commit and the release notes talk about introducing `MicrogridConfig` class which is already existing and already does what you describe as new.

cwasicki:
> I am confused with the second commit message

Indeed, sorry, that was the wrong message, fixed in the commit message, release notes and PR description.

cyiallou (approved):
Looks good to me

PR #72, opened 2025-05-02
Fix: Handle KeyError in StatisticalPlotStyle duration calculation

cyiallou:
Catch `KeyError` when `data_index` is empty or has one entry, causing `mode()` to return an empty `Series` during duration calculation in `StatisticalPlotStyle.get_plot_styles()`. Set duration to `None` to allow plot style generation to proceed without error.


flora-hofmann-frequenz (dismissed):

cyiallou:
Requires reapproval. Dealt with a conflict in the release notes file.

flora-hofmann-frequenz (approved):

PR #73, opened 2025-05-02
fix(models): pad predictions to match index length in prepare_prediction_models

cyiallou:
Fix a bug in `prepare_prediction_models` where model outputs could mismatch the expected index length when the available data is smaller than the configured moving average window size.

Instead of slicing and risking index errors, predictions are now aligned to the reference index by right-aligning predictions and padding missing entries with NaNs if necessary.

This makes the model preparation robust even with minimal data and prevents silent failures or incorrect series construction.


flora-hofmann-frequenz (approved):
LGTM

PR #75, opened 2025-05-05
Add simple NoDataAvailableError exception for missing data handling

cyiallou:
- Introduce a lightweight `NoDataAvailableError` exception to handle cases where no data is available for analysis or plotting.
- Add a check after data retrieval in `run_workflow()` to raise a `NoDataAvailableError` if reporting data is empty. This prevents proceeding with empty datasets and ensures graceful failure.
- Skip microgrid if no data is available during workflow.
- Skip real-time plot if higher-frequency data is missing.
- Skip weather-based model when weather data is missing.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,4 +13,4 @@
 
 ## Bug Fixes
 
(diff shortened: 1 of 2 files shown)


flora-hofmann-frequenz (approved):
LGTM

PR #82, opened 2025-05-09
Fix: Include current-day production

cyiallou:
Fixed a bug introduced in version `v0.5.2` where the microgrid tabular overview excluded current-day values from the total production for the past 30 and 365 days, causing inconsistencies with metrics that included today's data.


cyiallou:
The bug was introduced in #74

flora-hofmann-frequenz (dismissed):
LGTM

cyiallou:
Requires reapproval. Dealt with a conflict in the release notes file.

flora-hofmann-frequenz (approved):

PR #85, opened 2025-05-15
Major upgrade of numpy dependency to v2

cwasicki:
Also relaxed floating point requirements to avoid type checking errors.


cyiallou (approved):
looks good to me

PR #102, opened 2025-06-12
Add data_processing module to reporting

flora-hofmann-frequenz:
An effort to take out all data processing functions from the Reporting NB in Deepnote.

Example [NB](https://deepnote.com/workspace/Frequenz-30b27450-f7a3-4895-98f0-b368eca2b9b0/project/V3neustrom-Reporting-Notebook-Duplicate-d9455000-59c8-4b79-a7cc-59a1fd55952b/data_processing.py)

### DIFF
--- a/src/frequenz/lib/notebooks/reporting/data_processing.py
+++ b/src/frequenz/lib/notebooks/reporting/data_processing.py
@@ -1,7 +1,45 @@
 # License: MIT
 # Copyright © 2025 Frequenz Energy-as-a-Service GmbH
 
-"""Data processing functions for the reporting module."""
+"""Data processing functions for the reporting module.
+
+Overview:
+---------
+This module contains a series of data transformation functions used to generate
+energy reports from microgrid component data, such as PV (photovoltaic), battery,
+and grid consumption metrics. These functions are typically executed in a specific
+order within the reporting notebook. The output of one function is often used as
+input to the next, forming a processing pipeline.
+
+The functions handle:
+- Timezone normalization
+- Data enrichment (e.g., PV metrics, grid net usage)
+- Column renaming based on component configuration
+- Aggregation and summarization of energy data
+- Generation of overview tables and analysis-ready DataFrames
+
+Assumptions and Requirements:
+-----------------------------
+- Input `df` must contain at least the columns: `"timestamp"`, `"grid"`
+- Additional columns like `"pv_neg"` and `"battery_pos"` are required
+  for PV and battery metrics.
+- Timestamps must be in datetime format; timezone-naive timestamps
+  are assumed to be in UTC.
+- Component configuration `mcfg` must implement
+  `component_type_ids(...)` returning a list of IDs.
+- `component_types` is a list containing any of: `"grid"`, `"consumption"`,
+  `"pv"`, `"battery"`, `"chp"`, `"ev"`.
+
+Output:
+-------
+Most functions return either:
+- A modified `pd.DataFrame` with new or renamed columns,
+- A summary `dict` of computed statistics,
+- Or a reshaped long-format DataFrame for visual inspection or plotting.
+
+This modular pipeline ensures flexibility while maintaining clear structure
+for preparing reproducible, component-aware energy reporting.
+"""
 
 from datetime import datetime
 from typing import Any, Dict, List, Tuple, Union
@@ -9,6 +47,32 @@ from zoneinfo import ZoneInfo
 
 import numpy as np
 import pandas as pd
+
+
+def _get_rename_map(component_types: List[str]) -> Dict[str, str]:
+    """Return a mapping from raw column names to human-readable German names."""
+    rename_map: Dict[str, str] = {
+        "timestamp": "Zeitpunkt",
+        "grid": "Netzanschluss",
+        "consumption": "Brutto Gesamtverbrauch",
+    }
+
+    if "battery" in component_types:
+        rename_map["battery"] = "Batterie Durchsatz"
+
+    if "pv" in component_types:
+        rename_map.update(
+            {
+                "pv": "PV Durchsatz",
+                "pv_prod": "PV Produktion",
+                "pv_self": "PV Eigenverbrauch",
+                "pv_bat": "PV in Batterie",
+                "pv_feedin": "PV Einspeisung",
+                "pv_self_consumption_share": "PV Eigenverbrauchsanteil",
+            }
+        )
+
+    return rename_map
 
 
 def convert_timezone(df: pd.DataFrame) -> pd.DataFrame:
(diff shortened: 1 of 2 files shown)


cwasicki:
> An effort to take out all data processing functions from the Reporting NB in Deepnote.

Are these all copied as-is from the reporting notebook?

cyiallou:
> Are these all copied as-is from the reporting notebook?

If so, do you plan to clean them up in this PR, or in a follow-up? Just asking because I noticed a few spots that could probably use a bit of tidying.

flora-hofmann-frequenz:
> If so, do you plan to clean them up in this PR, or in a follow-up? Just asking because I noticed a few spots that could probably use a bit of tidying.

Happy to clean either in this PR or in the next one. What would you recommend?

They are not exactly as in the reporting notebook at most blocks in the notebook are not even functions yet and have been add by many different programmers (Malte, Noah, me...). As a first step I turned all of them into functions and made sure the notebook runs smoothly (with testing of different use cases) using them as they are here now.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,335 @@
...
+    assert "timestamp" in df.columns, df
+    if df["timestamp"].dt.tz is None:
+        df["timestamp"] = df["timestamp"].dt.tz_localize("UTC")
+    df["timestamp"] = df["timestamp"].dt.tz_convert("Europe/Berlin")
+    return df
+
+
+def process_grid_data(df: pd.DataFrame) -> pd.DataFrame:
+    """Add 'Netzbezug' column for positive grid consumption."""
+    df["Netzbezug"] = df["Netzanschluss"].apply(lambda x: x if x > 0 else 0)
```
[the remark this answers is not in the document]
cyiallou: I agree. I think `df["Netzanschluss"].clip(lower=0)` is faster. I think it's also more readable.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,335 @@
...
+
+def rename_component_columns(
+    df: pd.DataFrame, component_types: List[str], mcfg: Any
+) -> pd.DataFrame:
+    """Rename component columns based on configuration."""
+    single_comp = [col for col in df.columns if col.isdigit()]
+    rename_comp: Dict[str, str] = {}
+    if "battery" in component_types:
+        # pylint: disable=protected-access
+        battery_ids = {str(i) for i in mcfg._component_types_cfg["battery"].meter}
```
cyiallou: I think `mcfg.component_type_ids(component_type="battery", component_category="meter")` should give you the same result as `mcfg._component_types_cfg["battery"].meter`. Former is cleaner in my opinion as it avoids the pylint hint.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,335 @@
...
+        # pylint: enable=protected-access
+        rename_comp.update(
+            {col: f"Batterie #{col}" for col in single_comp if col in battery_ids}
+        )
+    if "pv" in component_types:
+        # pylint: disable=protected-access
+        pv_ids = {str(i) for i in mcfg._component_types_cfg["pv"].meter}
```
cyiallou: Similar to my previous comment.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,335 @@
...
+                overview_df[column] = np.nan
+    return overview_df
+
+
+def print_pv_sums(
+    master_df: pd.DataFrame, resolution: pd.Timedelta, pv_columns: List[str]
+) -> None:
+    """Print formatted sums for each PV column."""
+    for pv in pv_columns:
+        pv_sum = round(master_df[pv].sum() * (resolution.seconds / 3600) * -1, 2)
```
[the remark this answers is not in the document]
cyiallou: This suggestion makes sense.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,335 @@
...
+        formatted_sum = (
+            f"{pv_sum:,.2f}".replace(",", "X").replace(".", ",").replace("X", ".")
+        )
+        print(f"{pv:<7}:   {formatted_sum} kWh")
+
+
+def create_pv_analyse_df(
```
cyiallou, line 332: The `if` and `else` blocks share a lot of similar code which is also similar to `create_battery_analyse_df()`. Consider whether it can be abstracted and reused.

cyiallou (commented):
Thanks for the implementation. I have left a few comments mainly around consistency, maintainability, and minor architectural points. Nothing blocking, but worth addressing where possible I think. Some general comments:

- DataFrame mutation: Some functions modify the input DataFrame in place. Consider calling .copy() early or documenting the mutation clearly if this is intended. 
- Hardcoded column names: Strings like "PV Produktion" and "Netzanschluss" are repeated throughout. Suggest extracting them as constants to reduce errors and improve maintainability. There are more hardcoded strings throughout the code.
- Timezone inconsistency: convert_timezone() uses "Europe/Berlin", while compute_peak_usage() uses "CET". Consider unifying this with a shared constant.
- Docstrings are missing Args and Returns. Adding them would improve clarity.
- Consider whether adding tests for some functions makes sense.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,343 @@
...
+def create_master_dfs(
+    df: pd.DataFrame, component_types: List[str], mcfg: Any
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    """Create master DataFrame and renamed DataFrame from raw data."""
+    df = df.reset_index()
+    df = convert_timezone(df)
+    rename_map: Dict[str, str] = {
+        "timestamp": "Zeitpunkt",
+        "grid": "Netzanschluss",
+        "consumption": "Netto Gesamtverbrauch",
```
cwasicki: Why "Netto", isn't this "Brutto"?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,343 @@
...
+            str(i)
+            for i in mcfg.component_type_ids(
+                component_type="pv", component_category="meter"
+            )
+        }
+        rename_comp.update({col: f"PV #{col}" for col in single_comp if col in pv_ids})
+    return df.rename(columns=rename_comp)
+
+
+def create_master_dfs(
```
cwasicki: Let's take a more descriptive name ;)

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,343 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Data processing functions for the reporting module."""
```
cwasicki: To me it's not really clear how these functions are used. Should they run in a particular order and the output of one function will be input to another? Most functions here assume some particular columns to exist on the input df. From the docs it's not really clear what they do and how they can be used, e.g. what requirements there are on the input data.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,343 @@
...
+    return df.rename(columns=rename_comp)
+
+
+def create_master_dfs(
+    df: pd.DataFrame, component_types: List[str], mcfg: Any
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    """Create master DataFrame and renamed DataFrame from raw data."""
+    df = df.reset_index()
+    df = convert_timezone(df)
+    rename_map: Dict[str, str] = {
```
cwasicki: We should have a rename mapping at the beginning of the file instead of having them spread over the code base.

flora-hofmann-frequenz:
Addressed in PR #106

PR #103, opened 2025-06-12
Support multiple config files in component data

cwasicki:


flora-hofmann-frequenz:
I think we need to run black here quickly, otherwise if I approve it will just be stale.

cwasicki:
Fixed

flora-hofmann-frequenz (approved):

PR #104, opened 2025-06-18
Refactor config file arguments of MicrogridData and MicrogridConfig classes

matthias-wende-frequenz:
The `load_configs` static method has been refactored to offer a more flexible and explicit way of providing configuration files.
    
Previously, the method accepted a *paths argument. The signature has been changed to now accept two distinct optional parameters:
    
* The first argument accepts a single configuration file or a list of files.
* The second argument accepts a directory.
    
For both parameters, inputs can be provided as either String or Path objects. It is mandatory to provide at least one of these arguments. The `MicrogridConfig` class will attempt to load all specified configuration files although the content of the loaded configuration files is not yet validated.
    
The `MicrogridData` class now accepts a dict of `MicrogridConfigs` as returned by `MicrogridData.load_configs` instead of a path.

### DIFF
--- a/src/frequenz/data/microgrid/component_data.py
+++ b/src/frequenz/data/microgrid/component_data.py
@@ -24,7 +24,7 @@ class MicrogridData:
         server_url: str,
         auth_key: str,
         sign_secret: str,
-        microgrid_config_path: str | list[str],
+        microgrid_configs: dict[str, MicrogridConfig] | None = None,
     ) -> None:
         """Initialize microgrid data.
 
@@ -32,22 +32,12 @@ class MicrogridData:
             server_url: URL of the reporting service.
             auth_key: Authentication key to the service.
             sign_secret: Secret for signing requests.
-            microgrid_config_path: Path(s) to the config file with microgrid components.
-
-        Raises:
-            ValueError: If no microgrid config path is provided.
+            microgrid_configs: MicrogridConfig dict mapping microgrid IDs to MicrogridConfigs.
         """
+        self._microgrid_configs = microgrid_configs
         self._client = ReportingApiClient(
             server_url=server_url, auth_key=auth_key, sign_secret=sign_secret
         )
-        paths = (
-            [microgrid_config_path]
-            if isinstance(microgrid_config_path, str)
-            else microgrid_config_path
-        )
-        if len(paths) < 1:
-            raise ValueError("At least one microgrid config path must be provided")
-        self._microgrid_configs = MicrogridConfig.load_configs(*paths)
 
     @property
     def microgrid_ids(self) -> list[str]:
@@ -97,6 +87,8 @@ class MicrogridData:
         formulas = {
             ctype: mcfg.formula(ctype, metric.upper()) for ctype in component_types
         }
+
+        logging.debug("Formulas: %s", formulas)
 
         metric_enum = Metric[metric.upper()]
         data = [
--- a/src/frequenz/data/microgrid/config.py
+++ b/src/frequenz/data/microgrid/config.py
@@ -3,10 +3,14 @@
 
 """Configuration for microgrids."""
 
+import logging
 import re
 import tomllib
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Any, Literal, cast, get_args
+
+_logger = logging.getLogger(__name__)
 
 ComponentType = Literal["grid", "pv", "battery", "consumption", "chp", "ev"]
 """Valid component types."""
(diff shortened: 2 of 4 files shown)


matthias-wende-frequenz:
~Will fix the the tests if we have a agree to proceed.~
Fixed

src/frequenz/data/microgrid/component_data.py
```diff
@@ -19,9 +20,7 @@
 class MicrogridData:
     """Fetch power data for component types of a microgrid."""
 
-    def __init__(
-        self, server_url: str, key: str, microgrid_config_path: str | list[str]
-    ) -> None:
+    def __init__(self, server_url: str, key: str, microgrid_config_path: Path) -> None:
```
cwasicki: I think we shouldn't even deal with the files here at all but pass in the dict of configs here.

matthias-wende-frequenz: Agreed, but we should pass the `Path` parameter instead of a dict and update the `MicrogridConfig.load_configs` code to handle a Path.

src/frequenz/data/microgrid/config.py
```diff
@@ -281,12 +285,12 @@ def formula(self, component_type: str, metric: str) -> str:
...
         formula = cfg.formula.get(metric)
         if not formula:
-            raise ValueError(f"{component_type} is missing formula for {metric}")
+            raise ValueError(f"{component_type} formula is missing for {metric}")
 
         return formula
 
     @staticmethod
-    def load_configs(*paths: str) -> dict[str, "MicrogridConfig"]:
+    def load_configs(*paths: Path) -> Dict[str, "MicrogridConfig"]:
```
cwasicki: Here it should also allow `str`

src/frequenz/data/microgrid/component_data.py
```diff
@@ -37,17 +38,24 @@ def __init__(
         Raises:
             ValueError: If no microgrid config path is provided.
         """
+        if isinstance(microgrid_config_path, str):
```
cwasicki: This would be needed in the config loader in the other module.

src/frequenz/data/microgrid/component_data.py
```diff
@@ -37,17 +38,24 @@ def __init__(
...
-        paths = (
-            [microgrid_config_path]
-            if isinstance(microgrid_config_path, str)
-            else microgrid_config_path
-        )
-        if len(paths) < 1:
-            raise ValueError("At least one microgrid config path must be provided")
-        self._microgrid_configs = MicrogridConfig.load_configs(*paths)
+
+        if microgrid_config_path.is_dir():
```
cwasicki: I wouldn't add this as it can result in other config toml files in the same directory being parsed incorrectly.

src/frequenz/data/microgrid/component_data.py
```diff
@@ -37,17 +38,24 @@ def __init__(
...
+            config_files = Path(microgrid_config_path).glob("*.toml")
+            self._microgrid_configs = MicrogridConfig.load_configs(*config_files)
+        elif microgrid_config_path.is_file():
```
cwasicki: This section also in the config loader below.

src/frequenz/data/microgrid/component_data.py
```diff
@@ -24,7 +25,7 @@ def __init__(
         server_url: str,
         auth_key: str,
         sign_secret: str,
-        microgrid_config_path: str | list[str],
+        microgrid_config_path: str | Path,
```
cwasicki: Eventually we wanted to replace this with passing in the dict of configs directly, so the logic below should be in the config loader.

cwasicki: Not sure if you want to make this part of this PR already though.

src/frequenz/data/microgrid/config.py
matthias-wende-frequenz: Yes, if anyone is using the config module without the `MicrogridData` class. Do you know if that's the case?

src/frequenz/data/microgrid/component_data.py
matthias-wende-frequenz: I'd rather prefer to add config file validation, which has also other benefits.

src/frequenz/data/microgrid/config.py
cwasicki: Yes, there are various cases.

src/frequenz/data/microgrid/config.py
```diff
@@ -285,23 +289,73 @@ def formula(self, component_type: str, metric: str) -> str:
 
         return formula
 
+    # move the code from the other class here
```
cwasicki: Remove

cwasicki (commented):
LGTM, just minor comment

src/frequenz/data/microgrid/config.py
matthias-wende-frequenz: fixed

cwasicki (dismissed):
@cyiallou Are you also fine with this change?

cyiallou (dismissed):
Yes I'm also fine with these changes. Thanks Matthias

cyiallou:
it needs a rebase. I just merged another PR

matthias-wende-frequenz:
@cyiallou @cwasicki needs to be re-approved after the rebase ;).

cyiallou (approved):

PR #106, opened 2025-06-20
Add refactored data processing module to reporting

flora-hofmann-frequenz:
@cyiallou & @cwasicki - this PR takes in most of Costa's recommendations from PR #102 

Can you review this one first, and then we decide if we should still look at #102?!

### DIFF
--- a/src/frequenz/lib/notebooks/reporting/data_processing.py
+++ b/src/frequenz/lib/notebooks/reporting/data_processing.py
@@ -1,0 +1,361 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Data processing utilities for microgrid energy reporting.
+
+This module provides a set of functions for processing, enriching, and analyzing
+time-series energy data from microgrid systems. It focuses on preparing data for
+PV (photovoltaic), battery, and grid energy flows, transforming it into a consistent
+structure for visualization, reporting, and analysis.
+
+Features
+--------
+- Enriches raw energy data with derived columns such as:
+  - PV production, self-consumption, feed-in, and battery charging.
+  - Net grid import and PV self-consumption share.
+- Handles time zone localization and conversion to Europe/Berlin.
+- Dynamically renames columns to more descriptive names, including
+  mapping component IDs (e.g., "PV #1", "Batterie #2").
+- Provides summary energy mix breakdowns (PV vs grid) in kWh, % share, and average kW.
+- Prepares tailored DataFrames for PV and battery analysis, supporting flexible
+  filtering by component.
+
+Main Functions
+--------------
+- `transform_energy_dataframe(df, component_types, mcfg)`:
+    Transforms a raw DataFrame with energy metrics into an enriched,
+    user-friendly format, adding PV, battery, and grid metrics.
+
+- `compute_power_df(main_df, resolution)`:
+    Computes total energy drawn from PV and grid sources over the given resolution,
+    returning a summary DataFrame with kWh, percentage, and average kW.
+
+- `print_pv_sums(main_df, resolution)`:
+    Prints total PV feed-in sums for each individual PV component
+    in a localized numeric format.
+
+- `create_pv_analysis_df(main_df, pv_filter, pvgrid_filter, pv_grid_filter_options)`:
+    Generates a DataFrame for PV analysis based on selected PV components
+    and whether to analyze PV alone, grid alone, or a grid/PV split.
+
+- `create_battery_analysis_df(main_df, bat_filter)`:
+    Creates a DataFrame for analyzing battery throughput, reshaping
+    it to long format for multi-battery analysis.
+
+Usage
+-----
+Typical usage involves:
+1. Loading a raw DataFrame with time-indexed energy measurements.
+2. Calling `transform_energy_dataframe` to process and enrich it.
+3. Using the resulting DataFrames to generate summaries,
+   for example with `compute_power_df`, `create_pv_analysis_df`, or
+   `create_battery_analysis_df` for visualization.
+"""
+
+from typing import Any, Dict, Iterable, List, Tuple, Union
+
+import pandas as pd
+
+# Constants
+TZ_NAME = "Europe/Berlin"
+COLUMN_TIMESTAMP = "timestamp"
+COLUMN_TIMESTAMP_NAMED = "Zeitpunkt"
+COLUMN_GRID = "grid"
+COLUMN_GRID_NAMED = "Netzanschluss"
+COLUMN_NET_IMPORT = "Netzbezug"
+COLUMN_CONSUMPTION = "consumption"
+COLUMN_CONSUMPTION_NAMED = "Brutto Gesamtverbrauch"
+COLUMN_BATTERY = "battery"
+COLUMN_BATTERY_POS = "battery_pos"
+COLUMN_BATTERY_NAMED = "Batterie Durchsatz"
+COLUMN_PV = "pv"
+COLUMN_PV_PROD = "PV Produktion"
+COLUMN_PV_NEG = "pv_neg"
+COLUMN_PV_EXCESS = "pv_excess"
+COLUMN_PV_FEEDIN = "PV Einspeisung"
+COLUMN_PV_SELF = "PV Eigenverbrauch"
+COLUMN_PV_BAT = "pv_bat"
+COLUMN_PV_IN_BAT = "PV in Batterie"
+COLUMN_PV_SHARE = "PV Eigenverbrauchsanteil"
+COLUMN_PV_THROUGHPUT = "PV Durchsatz"
+
+
+def transform_energy_dataframe(
+    df: pd.DataFrame,
+    component_types: List[str],
+    mcfg: Any,
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    """Transform and enrich energy dataframe.
+
+    This function processes a raw DataFrame containing energy metrics,
+    adding derived columns for PV production, battery throughput, and grid metrics.
+
+    Args:
+        df: Raw DataFrame with energy metrics, expected to have a datetime index.
+        component_types: List of component types present in the DataFrame (e.g., ["pv", "battery"]).
+        mcfg: Microgrid configuration object providing component type IDs and other metadata.
+
+    Returns:
+        A tuple containing:
+        - `main_df`: A DataFrame with main columns for visualization and reporting.
+        - `df_renamed`: A fully enriched DataFrame.
+    """
+    # Ensure the DataFrame has a datetime index
+    df = df.reset_index()
+
+    # Enrich with PV-related columns
+    if "pv" in component_types:
+        df[COLUMN_PV_PROD] = -df.get(COLUMN_PV_NEG, 0)
+        df[COLUMN_PV_EXCESS] = (df[COLUMN_PV_PROD] - df[COLUMN_CONSUMPTION]).clip(
+            lower=0
+        )
+
+        if "battery" in component_types:
+            df[COLUMN_PV_IN_BAT] = df[[COLUMN_PV_EXCESS, COLUMN_BATTERY_POS]].min(
+                axis=1
+            )
+        else:
+            df[COLUMN_PV_IN_BAT] = 0
+
+        df[COLUMN_PV_FEEDIN] = df[COLUMN_PV_EXCESS] - df[COLUMN_PV_IN_BAT]
+        df[COLUMN_PV_SELF] = (df[COLUMN_PV_PROD] - df[COLUMN_PV_EXCESS]).clip(lower=0)
+        df[COLUMN_PV_SHARE] = df[COLUMN_PV_SELF] / df[COLUMN_CONSUMPTION].replace(
+            0, pd.NA
+        )
+
+    # Convert timestamp to Berlin time
+    if df[COLUMN_TIMESTAMP].dt.tz is None:
+        df[COLUMN_TIMESTAMP] = df[COLUMN_TIMESTAMP].dt.tz_localize("UTC")
+    df[COLUMN_TIMESTAMP] = df[COLUMN_TIMESTAMP].dt.tz_convert(TZ_NAME)
+
+    # Basic renaming
+    rename_map: Dict[str, str] = {
+        COLUMN_TIMESTAMP: COLUMN_TIMESTAMP_NAMED,
+        COLUMN_GRID: COLUMN_GRID_NAMED,
+        COLUMN_CONSUMPTION: COLUMN_CONSUMPTION_NAMED,
+    }
+
+    if "battery" in component_types:
+        rename_map[COLUMN_BATTERY] = COLUMN_BATTERY_NAMED
+
+    if "pv" in component_types:
+        rename_map.update(
+            {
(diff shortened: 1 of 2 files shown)


src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    df = df.copy()
+    rename_map = _get_rename_map(component_types)
+
+    single_comp = [col for col in df.columns if col.isdigit()]
+    if "battery" in component_types:
+        battery_ids = {
+            str(i)
+            for i in mcfg.component_type_ids(
+                component_type="battery", component_category="meter"
```
cyiallou: Is it necessary to rely only on "meter" being available in the microgrid config file or could it work with "inverter" too?  This is a bit risky because by specifying the `component_category` to fetch, the returned value would be an empty list if "meter" is not found. If you don't specify the `component_category` the default logic will be applied which is first return meter IDs --> fallback to inverter IDs --> fallback to component IDs.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,251 @@
...
+        }
+        rename_map.update(
+            {col: f"Batterie #{col}" for col in single_comp if col in battery_ids}
+        )
+
+    if "pv" in component_types:
+        pv_ids = {
+            str(i)
+            for i in mcfg.component_type_ids(
+                component_type="pv", component_category="meter"
```
cyiallou: Same as in the previous comment. It's not an issue though given that we are taking care of these config files internally and so we can make sure they are correct for this purpose.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,251 @@
...
+    """Return a mapping from raw column names to human-readable German names.
+
+    Args:
+        component_types: List of component types such as 'pv', 'battery'.
+
+    Returns:
+        A dictionary mapping internal column names to localized names.
+    """
+    rename_map: Dict[str, str] = {
+        COLUMN_TIMESTAMP: "Zeitpunkt",
```
cyiallou: Any particular reason why this is not a constant? It's used in two places.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,251 @@
...
+        COLUMN_CONSUMPTION: COLUMN_CONSUMPTION_NAMED,
+    }
+
+    if "battery" in component_types:
+        rename_map[COLUMN_BATTERY] = COLUMN_BATTERY_NAMED
+
+    if "pv" in component_types:
+        rename_map.update(
+            {
+                "pv": "PV Durchsatz",
```
cyiallou: Same with my previous comment.

cyiallou (dismissed):
Thanks Flora. It looks good to me. I only have a couple of minor comments.

cyiallou:
Oh yes, I forgot about the CI checks. Those can be easily fixed I guess.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,252 @@
...
+    Returns:
+        A dictionary mapping internal column names to localized names.
+    """
+    rename_map: Dict[str, str] = {
+        COLUMN_TIMESTAMP: "Zeitpunkt",
+        COLUMN_GRID: COLUMN_GRID_NAMED,
+        COLUMN_CONSUMPTION: COLUMN_CONSUMPTION_NAMED,
+    }
+
+    if "battery" in component_types:
```
cwasicki, line 138: I don't think you need the `if` clauses for component types.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,252 @@
...
+- PV and battery metrics require additional fields
+  (e.g., `"pv_neg"`, `"battery_pos"`)
+- The timestamp column must be timezone-aware or UTC-naive
+- A configuration object (`mcfg`) must provide `component_type_ids(...)`
+- `component_types` must be a list containing any of:
+  `"grid"`, `"consumption"`, `"pv"`, `"battery"`, `"chp"`, `"ev"`
+
+Outputs:
+--------
+Each function returns one of:
```
cwasicki: Can't we have a single function here that prepares a single data frame with all columns that are required for the notebook? I don't understand why we have so many functions that are called from the notebook.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,252 @@
...
+    df = df.reset_index(drop=True)
+    df = convert_timezone(df)
+    df_renamed = apply_renaming(df, component_types, mcfg)
+    df_renamed = process_grid_data(df_renamed)
+
+    master_df = df_renamed[_get_master_columns(df_renamed.columns, component_types)]
+    return master_df, df_renamed
+
+
+def _get_master_columns(
```
cwasicki: I am still very unhappy about the `master` terminology here and in other places. I

flora-hofmann-frequenz: :-D good point - I'll take it out...

src/frequenz/lib/notebooks/reporting/data_processing.py
flora-hofmann-frequenz: Good point. We had lots and lots of discussions in the first Borna (mid13) reportings, cause inverter and meter data do not 100% match up (mostly due different resistence), but having a fallback without comparison is a good idea.

src/frequenz/lib/notebooks/reporting/data_processing.py
flora-hofmann-frequenz: There are customer who do not have a battery or pv and then I do not want the specific metrics or graphs to be displayed.

src/frequenz/lib/notebooks/reporting/data_processing.py
flora-hofmann-frequenz: It comes from the reporting API as "timestamp", but then we use it as "Zeitpunkt" in the plotter. I will change that in future.

flora-hofmann-frequenz:
Sorry this one took so long, but I went back to the drawing board: Now its mostly one function that enriches the output from the reporting API. The other functions that are added help with all the different configuration that we have between the different customers set ups, so I would like to keep them in this module (rather than having the notebook too convoluted). I tried to include all feedback give before here and in PR #102.

src/frequenz/lib/notebooks/reporting/data_processing.py
cwasicki: But this is just the rename map, it only renames columns that exist. You can define the full mapping independent of the data you select.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+COLUMN_PV_BAT = "pv_bat"
+COLUMN_PV_IN_BAT = "PV in Batterie"
+COLUMN_PV_SHARE = "PV Eigenverbrauchsanteil"
+COLUMN_PV_THROUGHPUT = "PV Durchsatz"
+
+
+def transform_energy_dataframe(
+    df: pd.DataFrame,
+    component_types: List[str],
+    mcfg: Any,
```
matthias-wende-frequenz, line 86: You shouldn't use the `Any` type annotation here. You can use the actual type, `MicrogridConfig` instead.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    """Transform and enrich energy dataframe.
+
+    This function processes a raw DataFrame containing energy metrics,
+    adding derived columns for PV production, battery throughput, and grid metrics.
+
+    Args:
+        df: Raw DataFrame with energy metrics, expected to have a datetime index.
```
cwasicki, line 94: What are the input columns required on the input data?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+                base.update({COLUMN_PV_IN_BAT, COLUMN_PV_SHARE})
+
+        # Add individual component columns like "PV #1", "Batterie #3", etc.
+        base.update({col for col in columns if "#" in col})
+
+        return [col for col in columns if col in base]
+
+    main_df = df_renamed[_get_main_columns(df_renamed.columns, component_types)]
+
+    return main_df, df_renamed
```
cwasicki, line 207: Why do we need two data frames as ouput?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+        else:
+            df[COLUMN_PV_IN_BAT] = 0
+
+        df[COLUMN_PV_FEEDIN] = df[COLUMN_PV_EXCESS] - df[COLUMN_PV_IN_BAT]
+        df[COLUMN_PV_SELF] = (df[COLUMN_PV_PROD] - df[COLUMN_PV_EXCESS]).clip(lower=0)
+        df[COLUMN_PV_SHARE] = df[COLUMN_PV_SELF] / df[COLUMN_CONSUMPTION].replace(
+            0, pd.NA
+        )
+
+    # Convert timestamp to Berlin time
```
cwasicki, line 126: Would move this to the beginning of this function.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+
+        return [col for col in columns if col in base]
+
+    main_df = df_renamed[_get_main_columns(df_renamed.columns, component_types)]
+
+    return main_df, df_renamed
+
+
+def compute_power_df(
+    main_df: pd.DataFrame, resolution: Union[str, pd.Timedelta]
```
cwasicki, line 211: `str | pd.Timedelta`

cwasicki: Why not a `datetime.timedelta` which we usually use?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+    Args:
+        main_df: DataFrame containing PV and grid data.
+        pv_filter: List of PV components to include (e.g., ["1", "2"] or ["Alle"]).
+        pvgrid_filter: Filter option for PV and grid analysis (e.g., "PV", "Grid", "PV + Grid").
+        pv_grid_filter_options: List of available filter options for PV and grid.
+    Returns:
+        A DataFrame with PV feed-in data, reshaped for analysis.
+    """
+    # Case 1: Only PV
+    if pvgrid_filter == pv_grid_filter_options[1]:
```
cwasicki, line 291: This should check for the strings. Otherwise you rely on something passed in which you cannot control.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+def create_pv_analysis_df(
+    main_df: pd.DataFrame,
+    pv_filter: List[str],
+    pvgrid_filter: str,
+    pv_grid_filter_options: List[str],
+) -> pd.DataFrame:
+    """Create a DataFrame for PV analysis based on selected filters.
+
+    Args:
+        main_df: DataFrame containing PV and grid data.
```
cwasicki, line 283: Which columns?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+    # Case 1: Only PV
+    if pvgrid_filter == pv_grid_filter_options[1]:
+        pv_columns = (
+            [col for col in main_df.columns if "PV #" in col]
+            if "Alle" in pv_filter
+            else [f"PV {pv}" for pv in pv_filter]
+        )
+        df = main_df[[COLUMN_TIMESTAMP_NAMED] + pv_columns].copy()
+        df = df.melt(
```
cwasicki, line 298: Why is it converted to long format?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+    pv_filter: List[str],
+    pvgrid_filter: str,
+    pv_grid_filter_options: List[str],
+) -> pd.DataFrame:
+    """Create a DataFrame for PV analysis based on selected filters.
+
+    Args:
+        main_df: DataFrame containing PV and grid data.
+        pv_filter: List of PV components to include (e.g., ["1", "2"] or ["Alle"]).
+        pvgrid_filter: Filter option for PV and grid analysis (e.g., "PV", "Grid", "PV + Grid").
```
cwasicki, line 285: What are these three modes?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+            else [f"PV {pv}" for pv in pv_filter]
+        )
+        df = main_df[[COLUMN_TIMESTAMP_NAMED, COLUMN_GRID_NAMED] + pv_columns].copy()
+        df = df.melt(
+            id_vars=[COLUMN_TIMESTAMP_NAMED, COLUMN_GRID_NAMED],
+            value_vars=pv_columns,
+            var_name="PV",
+            value_name=COLUMN_PV_FEEDIN,
+        )
+        df[COLUMN_GRID_NAMED] /= len(pv_columns)
```
cwasicki, line 326: What is this calculating?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+
+    for pv in pv_columns:
+        pv_sum = round(main_df[pv].sum() * resolution * -1, 2)
+        formatted_sum = (
+            f"{pv_sum:,.2f}".replace(",", "X").replace(".", ",").replace("X", ".")
+        )
+        print(f"{pv:<7}:   {formatted_sum} kWh")
+
+
+def create_pv_analysis_df(
```
cwasicki, line 274: I don't really get what this and the next function are required for.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,361 @@
...
+COLUMN_PV_EXCESS = "pv_excess"
+COLUMN_PV_FEEDIN = "PV Einspeisung"
+COLUMN_PV_SELF = "PV Eigenverbrauch"
+COLUMN_PV_BAT = "pv_bat"
+COLUMN_PV_IN_BAT = "PV in Batterie"
+COLUMN_PV_SHARE = "PV Eigenverbrauchsanteil"
+COLUMN_PV_THROUGHPUT = "PV Durchsatz"
+
+
+def transform_energy_dataframe(
```
cwasicki, line 83: For this main function it would be good to have a test.

PR #112, opened 2025-06-27
Fix hardcoded feature name

cyiallou:


cyiallou:
I will create a patch release after this PR. I missed this line in the previous release. In any case, the `config.py` requires heavy refactoring to properly define the configurations as dataclasses and pass in the entire code base which would avoid such cases too.

flora-hofmann-frequenz (approved):

PR #122, opened 2025-07-03
Update Solar Maintenance notebook

cyiallou:
You can now process multiple microgrid TOML files and choose between "inverter" and "meter" component types for PV data fetching. Added a new input block to enable the latter. Also updated the notebook instructions to describe the new input parameter.


flora-hofmann-frequenz (approved):
LGTM!

PR #129, opened 2025-07-11
Update Alert Notebook

cyiallou:
- Update the introduction in the instructions.
- Utilise the updated `MicrogridConfig` class to simplify the toml
  file loading.
- Utilise the new `state_analysis` module for detecting and analysing
  component state transitions and alerts from reporting data. This
  simplifies the notebook by removing a now obsolete code.


cyiallou:
please also double check the German translation

cwasicki:
There is still quite a lot of code in the notebook, which would be better moved to the lib. Also we should aim for having only a single library (notebooks) that the users need to worry about, not 3.

cwasicki:
> please also double check the German translation

Translation LGTM

cyiallou:
> There is still quite a lot of code in the notebook, which would be better moved to the lib. 

There is some code that will be deleted once this PR is merged https://github.com/frequenz-floss/frequenz-reporting-python/pull/58

I will review other parts of the notebook to see what can be moved to the library. Do  you see specific parts that should be moved already?

> Also we should aim for having only a single library (notebooks) that the users need to worry about, not 3.

You are right. Just realised that `client-common` can be removed from the notebook because it is already installed as a sub-dependency. The `frequenz-reporting-python` is needed because I am using the code from https://github.com/frequenz-floss/frequenz-reporting-python/pull/58 so this dependency can also be dropped if we move most of the `frequenz/reporting/_reporting.py` module (all code except `cumulative_energy()`) in `lib-notebooks`. What do you think?

cwasicki:
From my side it's fine to do this later. We should aim for notebooks that have just couple of lines for interaction with the user and display, and everything else in a python module.

cyiallou:
Updated @cwasicki

cwasicki (approved):

PR #130, opened 2025-07-14
Add transactional stateful data fetcher

matthias-wende-frequenz:
This class is a wrapper around the `MicrogridData` class to add state designed for continues data exports. It not only receives data, but also writes the latest data point to an on disc buffer. The timestamp in the buffer is then used as the start time for the next run.

Furthermore this module introduces a transactional approach to updating the on disc buffer to ensure data integrity:

* Buffer updates are written to a temporary file (e.g., `data.parquet.tmp`).
* On success: The temporary file is atomically renamed to its final path, committing the new buffer state.
* On failure: The temporary file is deleted, rolling back the change.

This ensures the unprocessed data will be fetched again on the next run.

### DIFF
--- a/src/frequenz/data/microgrid/_stateful_data_fetcher.py
+++ b/src/frequenz/data/microgrid/_stateful_data_fetcher.py
@@ -1,0 +1,274 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+"""A helper module to write data exports.
+
+This class is a wrapper around the `MicrogridData` class to add state
+designed for continuous data exports. It not only receives data, but also
+writes the latest data point to an on disc buffer. The timestamp in the
+buffer is then used as the start time for the next run.
+
+Furthermore this module introduces a transactional approach to updating
+the on disc buffer to ensure data integrity:
+
+- Buffer updates are written to a temporary file (e.g.,
+data.parquet.tmp).
+- On success: The temporary file is atomically renamed to its final
+path, committing the new buffer state.
+- On failure: The temporary file is deleted, rolling back the change.
+
+This ensures the unprocessed data will be fetched again on the next
+run.
+
+???+ example
+
+    ```python
+    REPORTING_API_KEY = "your_api_key"
+    REPORTING_API_SECRET = "your_api_secret"
+
+    mid = 12345
+    component_name = "battery"
+    metric = "active_power"
+
+    mg_config = MicrogridConfig.load_configs(
+        microgrid_config_dir=Path("/path/to/microgrid/configs")
+    )
+    mg_data = MicrogridData(
+        "https://reporting.example.com",
+        REPORTING_API_KEY,
+        REPORTING_API_SECRET,
+        mg_config,
+    )
+
+    data_fetcher = StatefulDataFetcher(
+        microgrid_data,
+        data_buffer_dir=Path("/path/to/data/buffer"),
+        resampling_period=timedelta(seconds=180),
+    ),
+
+    try:
+        data = await data_fetcher.receive_microgrid_data(
+            mid,
+            component_name,
+            metric,
+        )
+        _logger.debug(
+            "Data fetched for %s: %s", location.component.name, data
+        )
+    except ValueError as e:
+        _logger.error("Error fetching data for AKS ID %s: %s", aks_id, e)
+
+    # do sth with data
+    # ...
+
+    data_fetcher.commit()  # Commit the changes if successful
+    # data_fetcher.rollback()  # Rollback if there was an error
+    ```
+"""
+
+import logging
+import os
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+
+import pandas as pd
+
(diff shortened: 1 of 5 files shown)


src/frequenz/data/microgrid/stateful_data_fetcher.py
```diff
@@ -0,0 +1,196 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+"""A helper module to write data exports."""
```
matthias-wende-frequenz: Extend the documentation and add usage examples.

src/frequenz/data/microgrid/_stateful_data_fetcher.py
```diff
@@ -0,0 +1,266 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
nkrtzr: 2025

nkrtzr (changes requested):
Beside minor change all good

src/frequenz/data/microgrid/_stateful_data_fetcher.py
matthias-wende-frequenz: fixed

src/frequenz/data/microgrid/_stateful_data_fetcher.py
```diff
@@ -0,0 +1,266 @@
...
+        self._data_buffer = data_buffer_dir
+        self._resampling_period = resampling_period
+        self._end_time_delta = end_time_delta
+
+        self._data_buffer.mkdir(parents=True, exist_ok=True)
+
+        # Maps temp_path -> final_path for files that need to be committed
+        self._temp_files_to_commit: dict[Path, Path] = {}
+
+    def _generate_filename(
```
cwasicki, line 129: Could be `staticmethod`

cwasicki: Or could prepend the data_buffer.

src/frequenz/data/microgrid/_stateful_data_fetcher.py
```diff
@@ -0,0 +1,266 @@
...
+        end_time = now - self._end_time_delta
+        on_disk_buffer = pd.DataFrame()
+        start_time: datetime
+
+        try:
+            on_disk_buffer = pd.read_parquet(final_parquet_file)
+            if not on_disk_buffer.empty and isinstance(
+                on_disk_buffer.index, pd.DatetimeIndex
+            ):
+                last_timestamp = on_disk_buffer.index[-1].to_pydatetime()
```
cwasicki: You could use `index.max()` to not rely on sorting.

src/frequenz/data/microgrid/_stateful_data_fetcher.py
```diff
@@ -0,0 +1,266 @@
...
+                start_time = last_timestamp + timedelta(microseconds=1)
+                _logger.info("Read buffer. Last timestamp: %s.", last_timestamp)
+            else:
+                _logger.info("Buffer file is empty. Defaulting to last hour.")
+                start_time = now - timedelta(hours=1)
```
cwasicki: Does that mean on first start it only starts with the last hour? Maybe configurable?

src/frequenz/data/microgrid/_stateful_data_fetcher.py
```diff
@@ -0,0 +1,266 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+"""A helper module to write data exports.
+
+This class is a wrapper around the `MicrogridData` class to add state
+designed for continues data exports. It not only receives data, but also
```
cwasicki: continuous?

src/frequenz/data/microgrid/_stateful_data_fetcher.py
cwasicki: but since you sort it before writing it's fine

src/frequenz/data/microgrid/_stateful_data_fetcher.py
cwasicki: Or better an argument to the method. Because it depends on the user how often they call it.

cwasicki (dismissed):
LGTM, but I would give the user control about the start and end times of the request.

src/frequenz/data/microgrid/_stateful_data_fetcher.py
matthias-wende-frequenz: updated

matthias-wende-frequenz:
> but I would give the user control about the start and end times of the request.

fixed, such that the user can control the initial period. Having control over start time doesn't make sense is this is by design read from disk. Controlling the end time could be discussed if necessary. I've chosen to hide it from the user to avoid misconfiguration.

cwasicki (dismissed):

cwasicki (approved):

PR #146, opened 2025-08-28
feat: Add data processing module

Mohammad-Tayyab-Frequenz:


RELEASE_NOTES.md
```diff
@@ -16,3 +16,5 @@
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+
+* Add reusable, modular data processing functions
```
cwasicki: This needs to be more informative, e.g. that it affects reporting functions. The repo currently contains various unrelated modules.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+        "Energiebezug": [name for name, _ in rows],
+        "Energie [kWh]": [round(val, 2) for _, val in rows],
+        "Energie %": [round(val / denom * 100, 2) for _, val in rows],
+        "Energie [kW]": [round((val / hours) if hours else 0.0, 2) for _, val in rows],
+    }
+    return pd.DataFrame(data)
+
+
+def compute_pv_metrics(  # pylint: disable=too-many-locals
+    master_df: pd.DataFrame, resolution: timedelta, grid_consumption_sum: float
```
cwasicki: I would like to avoid the `master_df` name throughout this module. I know that it's taken from the notebook but I don't see any value in this over just calling it `df` or some other name that is actually descriptive.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+COLUMN_PV_FEEDIN = "PV Einspeisung"
+COLUMN_PV_SELF = "PV Eigenverbrauch"
+COLUMN_PV_BAT = "pv_bat"
+COLUMN_PV_IN_BAT = "PV in Batterie"
+COLUMN_PV_SHARE = "PV Eigenverbrauchsanteil"
+COLUMN_PV_THROUGHPUT = "PV Durchsatz"
+
+
+def compute_power_df(
+    df: pd.DataFrame, resolution: Union[str, pd.Timedelta]
```
cwasicki: Should be `str | pd.Timedelta`.

Also, do we need support for both?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Microgrid Reporting and Analysis Utilities.
```
cwasicki: Can you also add the notebook to see how the functions in this module are used?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+COLUMN_PV_EXCESS = "pv_excess"
+COLUMN_PV_FEEDIN = "PV Einspeisung"
+COLUMN_PV_SELF = "PV Eigenverbrauch"
+COLUMN_PV_BAT = "pv_bat"
+COLUMN_PV_IN_BAT = "PV in Batterie"
+COLUMN_PV_SHARE = "PV Eigenverbrauchsanteil"
+COLUMN_PV_THROUGHPUT = "PV Durchsatz"
+
+
+def compute_power_df(
```
cwasicki: This name doesn't describe what the function is doing.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+
+    if ts.dt.tz is None:
+        # Assume naïve datetimes are in `assume_tz`
+        ts = ts.dt.tz_localize(assume_tz)
+
+    df[column_timestamp] = ts.dt.tz_convert(target_tz)
+    return df
+
+
+def process_grid_data(
```
cwasicki: This function is overkill for a single line:
```
df[col_net_import] = df[col_grid].clip(lower=0)
```

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+    target_tz: str = "Europe/Berlin",
+    assume_tz: str = "UTC",
+) -> pd.DataFrame:
+    """
+    Convert the datetime column defined by ``column_timestamp`` to a target timezone.
+
+    If the column contains timezone-naive datetimes, they are first localized to
+    ``assume_tz`` before being converted to ``target_tz``.
+
+    Parameters
```
cwasicki: Wrong docstring style

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+    mcfg: MicrogridConfig,
+    column_battery: str = "battery",
+    column_pv: str = "pv",
+) -> Tuple[pd.DataFrame, List[str]]:
+    """
+    Rename numeric single-component columns to labeled names.
+
+    Numeric string column names like ``"14"`` are converted to
+    ``"Batterie #14"`` or ``"PV #14"`` based on ``mcfg._component_types_cfg``.
+    If an ID appears in both battery and PV sets, the PV label takes precedence.
```
cwasicki: An ID in both would be a fundamental problem in the component graph and better raise an error.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+        Configuration with ``_component_types_cfg`` mapping component types to a
+        ``meter`` iterable of numeric IDs.
+
+    Returns
+    -------
+    tuple[pd.DataFrame, list[str]]
+        DataFrame with renamed columns, and the list of applied labels in order.
+    """
+    # Numeric component columns present in df
+    single_components = [c for c in df.columns if isinstance(c, str) and c.isdigit()]
```
cwasicki: Could it also be an int?

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # From config (empty set if missing)
+    battery_components = set(
```
cwasicki: Could you use this instead? https://github.com/frequenz-floss/frequenz-lib-notebooks/blob/a33a55f57d5815f48eea2776fe064b1be675a656/src/frequenz/data/microgrid/config.py#L233

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+    )
+
+    pv_components = set(
+        map(str, getattr(mcfg._component_types_cfg.get(column_pv), "meter", []) or [])
+    )
+
+    rename = {}
+    rename.update(
+        {c: f"Batterie #{c}" for c in single_components if c in battery_components}
```
cwasicki: The `"Batterie"` part would better be configurable, also default to `"Battery"`.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+COLUMN_PV_PROD = "PV Produktion"
+COLUMN_PV_NEG = "pv_neg"
+COLUMN_PV_EXCESS = "pv_excess"
+COLUMN_PV_FEEDIN = "PV Einspeisung"
+COLUMN_PV_SELF = "PV Eigenverbrauch"
+COLUMN_PV_BAT = "pv_bat"
+COLUMN_PV_IN_BAT = "PV in Batterie"
+COLUMN_PV_SHARE = "PV Eigenverbrauchsanteil"
+COLUMN_PV_THROUGHPUT = "PV Durchsatz"
+
```
cwasicki, line 40: Better also have the data fetching part in this module, similar to what we do in the asset optimization module [here](https://github.com/frequenz-floss/frequenz-lib-notebooks/blob/a33a55f57d5815f48eea2776fe064b1be675a656/src/frequenz/lib/notebooks/reporting/asset_optimization/data.py#L53). In fact these two modules should be merged at some point.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
```
cwasicki, line 2: I had a couple of comments on https://github.com/frequenz-floss/frequenz-lib-notebooks/pull/106 already, some which apply here too, e.g.
* Without seeing the notebook I find it very hard to understand in which order the functions are expected to be called.
* Also there are quite a few functions here, are they really all needed in the notebook? I would try to expose as few functions as possible.
* There are no tests. I know it's difficult to test all of this, but some basic tests would be good. Can also be a follow-up PR.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+from frequenz.data.microgrid.config import MicrogridConfig
+from frequenz.lib.notebooks.reporting.utils.helpers import (
+    convert_timezone,
+    process_grid_data,
+    rename_single_components,
+)
+
+TZ_NAME = "Europe/Berlin"
+ASSUME_TZ = "UTC"
+COLUMN_TIMESTAMP = "timestamp"
```
cwasicki: I think these column names should be managed via a dictionary or some other approach. In the end we want the user to be able to change these names right?

src/frequenz/lib/notebooks/reporting/data_processing.py
cwasicki: And why not `datetime.timedelta`?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+      resolution: Row time resolution (e.g., ``"15min"`` or
+        ``pd.Timedelta("0 days 00:15:00")``).
+
+    Returns:
+      pd.DataFrame: A summary table with columns:
+        - ``Energiebezug``: Source label (``"PV"`` or ``"Netz"``).
+        - ``Energie [kWh]``: Total energy per source.
+        - ``Energie %``: Percentage share of total energy.
+        - ``Energie [kW]``: Average power over the interval.
+    """
```
cwasicki: This function looks complicated, isn't it just

```
E = df[[NET, SELF]].sum(axis=0) * resolution / timedelta(hours=1)
...
```

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+    resolution = pd.to_timedelta(resolution)
+    hours = resolution.total_seconds() / 3600.0
+    pv_kwh = 0.0
+    grid_kwh = 0.0
+
+    if COLUMN_NET_IMPORT in df.columns:
+        grid_kwh = round(df[COLUMN_NET_IMPORT].sum() * hours, 2)
```
cwasicki: I wouldn't do the rounding in the calculations here, but only have this in the notebook or optionally as an argument to be applied to the dataframe when returning.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+        return {}
+
+    step_hours = (
+        (resolution.total_seconds() / 3600.0)
+        if hasattr(resolution, "total_seconds")
+        else (resolution.seconds / 3600.0)
+    )
+
+    # German number formatter (e.g., 1,234.56 -> 1.234,56)
+    def fmt_de(x: float) -> str:
```
cwasicki: This should be a helper. Also only apply it optionally.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -0,0 +1,567 @@
...
+
+    # Keep only the number after "Batterie "
+    bat_analyse_df["Batterie"] = bat_analyse_df["Batterie"].str.replace(
+        "Batterie ", "", regex=False
+    )
+
+    return bat_analyse_df
+
+
+def add_pv_energy_flows(df: pd.DataFrame) -> pd.DataFrame:
```
cwasicki: If not necessary for performance reasons (which is not the case here), better not expose a function that modifies the data frame in place but always create a copy and modify that.

src/frequenz/lib/notebooks/reporting/data_processing.py
cwasicki: Also it looks like you have a set of input columns and output columns which you map below. That mapping should be configurable.

cwasicki (commented):
I think it would be good to have a separate module which clearly defines every new metric or KPI. These could be implemented as a function for each of these, which takes all relevant series as input arguments and returns the corresponding series or scalar, resp. 

Then the reporting module provides a single function that takes the data frame received from [this](https://github.com/frequenz-floss/frequenz-lib-notebooks/blob/a33a55f57d5815f48eea2776fe064b1be675a656/src/frequenz/data/microgrid/component_data.py#L188) as input and creates a data frame with new columns just using these functions. 

Similarly further KPIs or data frames which do not share the same time index can be derived from that new data frame.

What do you think @Mohammad-Tayyab-Frequenz?

RELEASE_NOTES.md
Mohammad-Tayyab-Frequenz: Sure, will change that.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Changed the name to energy_report_df now.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: The idea was to support both str and timedelta. Currently in notebooks, we convert resolution to timedelta. So, we can safely remove it from here I think.

Yeah, datetime.timedelta makes more sense in this case.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Sure, I will add the updated notebook also in the next commit.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Yup, changed to a more relevant function name.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Yea, the idea was to keep it separate in case we add any more processing in the future.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Yes, I will change the docstring of PR as well.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Got it. Will make that change.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: The dataframe that we receive from the API provide the column names in str format, but I think this condition is now redundant as I am already check if the column name is a digit. So, I will remove the first if condition.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Yeah, makes sense. Didn't know we had a direct function

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Sure.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Yeah, I would use that function directly in the notebook. Would make more sense than adding the same module in a different folder.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Understood. I’ll update this to expose only the necessary functions.

I would write another PR for tests.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Yeah, I picked this up setting from a previous PR. But do you think we should manage it using dictionary or a .yaml file?
.yaml file would be more manageable in the long run.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: For the computational part, it is similar. But I am creating a dataframe which has sum, % and energy in KW.
So, I can incorporate this in the original function.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Got it. Will change this.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Yeah, makes sense.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Sure, I will not expose it anymore. Also, create a copy within the function.

src/frequenz/lib/notebooks/reporting/data_processing.py
cwasicki: Up to you, let's try yaml then.

src/frequenz/lib/notebooks/reporting/data_processing.py
cwasicki: Ah, you want to create one function that covers both? Good idea!

src/frequenz/lib/notebooks/reporting/data_processing.py
cwasicki: Nevermind, just thought the whole function can be done simpler and more readable, but not super important for now.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
cwasicki: In fact I would see an argument in keeping it if it is renamed. Idea is also to have clearly defined metric definitions, so for this it would make sense. But if I read `process_grid_data` somewhere in the code I couldn't guess that it adds a grid consumption column.

Mohammad-Tayyab-Frequenz: Yes, the name does not make complete sense. I will change it.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Added the notebook in the examples folder.

PR #159, opened 2025-09-10
fix: handle empty DataFrame in solar notebook

cyiallou:
Prevents failure when the notebook processes an empty DataFrame. Follows the change made in commit df4e5a8


RELEASE_NOTES.md
```diff
@@ -15,4 +15,4 @@
 ## Bug Fixes
 
 * Replaces multiple duplicated plot functions with a single reusable one.
-* Handle empty weather/reporting dataframes gracefully to avoid transformation errors.
+* Handle empty weather/reporting dataframes gracefully to avoid transformation errors. The "Solar Maintenance" notebook is updated accordingly.
```
cwasicki, line 18: Can't this logic be part of the module instead of the notebook?

cwasicki (approved):

RELEASE_NOTES.md
cyiallou: The module returns a dataclass with the plot dataframes. These dataframes are dynamic and the plots generated by the module adjust accordingly.

The changes in this PR are specific to ensuring that Deepnote plots are generated correctly inside the notebook environment. Since this handling is environment-specific rather than part of the core functionality, I don’t think it belongs in the module itself. Keeping it in the notebook keeps the module clean and focused, while still making sure Deepnote users don’t run into issues.

cyiallou: I am merging the PR for now. We can discuss if you still have concerns about this

PR #160, opened 2025-09-12
Revert "Widen weather API client dependency range (#145)"

cyiallou:
This reverts commit 33fe0fe. `v0.3.0` of the weather client has a breaking
change which affects the module `frequenz.datasci.weather.weather_api.py`
In particular, the client method `hist_forecast_iterator` is replaced by
the `async` method `stream_historical_forecast`. Also, `v0.3.0` uses a
new version of the API, that the weather service doesn't support yet.
When the service supports it, the client dependency should be updated
to minimum `v0.3.0` due to the breaking change.

It also bumps `frequenz-client-reporting` to min `v0.19.0`

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* Revert back to `v0.2.3` of the weather-client until the weather service is able to support `v0.13` of the weather API.
 
 ## New Features
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
   "toml >= 0.10.2, < 0.11.0",
   "plotly >= 6.0.0, < 6.4.0",
   "kaleido >= 0.2.1, < 1.1.0",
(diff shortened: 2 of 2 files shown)


cyiallou:
@cwasicki please confirm the changes here because they revert the changes you made in #145

cwasicki:
> In particular, the client method hist_forecast_iterator is replaced by the async method stream_historical_forecast

Shouldn't it have failed the CI? Anyway, it's not nice but I don't have a better idea to handle this.

cwasicki (approved):

cyiallou:
> Shouldn't it have failed the CI?

I double-checked the [logs](https://github.com/frequenz-floss/frequenz-lib-notebooks/actions/runs/17130510799/job/48593208475) and CI did install `frequenz-client-weather==0.3.1`, but all checks still passed. Locally, when I ran `nox`, it immediately flagged the missing `hist_forecast_iterator`. I’m wondering if this is simply because we don’t have any tests for `weather_api.py`, so `pytest` never exercises that code path, and maybe `mypy` doesn’t catch it either (possibly due to type information not being strict enough).

Could that explain why CI didn’t complain even though the breakage is [real](https://github.com/frequenz-floss/frequenz-client-weather-python/pull/49/files)?

PR #164, opened 2025-09-24
feat: add a column mapper class

Mohammad-Tayyab-Frequenz:


src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
```diff
@@ -0,0 +1,90 @@
...
+  assume_tz: "UTC"
+
+columns:
+  timestamp:
+    raw: "timestamp"
+    display:
+      en: "Timestamp"
+      de: "Zeitpunkt"
+
+  # grid_consumption:
```
cwasicki: If it's not used this can be removed.

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
```diff
@@ -0,0 +1,90 @@
...
+  #   raw: "grid"
+  #   display:
+  #     en: "Grid Connection"
+  #     de: "Netzanschluss"
+
+  net_import:
+    raw: "Netzbezug"
```
cwasicki: Why a German term here?

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
```diff
@@ -0,0 +1,149 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Column mapping utilities for energy reporting.
+
+Provides the `ColumnMapper` dataclass to manage renaming between raw,
+canonical, and localized display column names. Supports loading schema
```
cwasicki: What are raw, canonical and display names? Where are these used?

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
```diff
@@ -0,0 +1,149 @@
...
+
+import pandas as pd
+import yaml
+
+
+@dataclass(frozen=True)
+class ColumnMapper:  # pylint: disable=too-many-instance-attributes
+    """Column schema with locale-aware display labels."""
+
+    version: int
```
cwasicki: These fields need docstrings

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
```diff
@@ -0,0 +1,90 @@
...
+    display:
+      en: "Grid Import"
+      de: "Netzbezug"
+
+  net_consumption:
+    raw: "consumption"
```
cwasicki: Why not net consumption?

cwasicki (changes requested):
Just from this PR it's not clear what this is about because it is missing PR and commit descriptions and the doc is not sufficient. This is missing a motivation why this class is needed (in the future) and how it is intended to be used.

I am also not sure about a couple of name matchings. This stresses the point that we need well-defined metric definitions. In fact I think we should add the module with the definitions of all of these metrics used here either before this PR or add them here.

cyiallou:
When I added EN/DE support for the solar notebook, I faced a similar challenge. At the time I went for a simple TranslationManager class with an internal dictionary of text + translations ([link](https://github.com/frequenz-floss/frequenz-lib-notebooks/blob/v0.x.x/src/frequenz/lib/notebooks/solar/maintenance/translator.py)). It’s admittedly not the most elegant solution, but it avoids introducing an extra dependency in this repo (which is itself consumed elsewhere). That trade-off was one of the reasons I kept it simple.

I see the YAML approach here is more structured (which is definitely a plus for maintainability). Just wanted to flag the existing `translator.py` so you can consider whether it makes sense to adapt/reuse it, or whether your case really needs a separate schema + class.

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
Mohammad-Tayyab-Frequenz: Sure, I will remove the current unused one.

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
Mohammad-Tayyab-Frequenz: Yes, previously we were creating netzbezug column but I have changed it to net_import now. I will change it here as well.

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
Mohammad-Tayyab-Frequenz: I will update this in the documentation. 

Raw - directly from reporting api
Canonical - used in our codebase
Display - To be displayed in the notebooks

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
Mohammad-Tayyab-Frequenz: Yes, I will add these.

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
Mohammad-Tayyab-Frequenz: I need to redefine the metrics once again so that correct term is used everywhere.

Mohammad-Tayyab-Frequenz:
The translator.py script looks well-structured and offers a wide range of functionalities. My current solution is more narrowly focused, handling only the translation of column display names into the required language. It might be worthwhile to combine both approaches and define a comprehensive configuration within a .yaml file. Do you think this would be a feasible path forward?

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
```diff
@@ -0,0 +1,84 @@
...
+      en: "Timestamp"
+      de: "Zeitpunkt"
+
+  grid_import:
+    raw: "grid_import"
+    display:
+      en: "Grid Import"
+      de: "Netzbezug"
+
+  brutto_consumption:
```
Mohammad-Tayyab-Frequenz: change to mid consumption

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+columns:
+  timestamp:
+    raw: "timestamp"
+    display:
+      en: "Timestamp"
+      de: "Zeitpunkt"
+
+  grid_import:
+    raw: "grid_import"
```
Mohammad-Tayyab-Frequenz: grid_import to grid load

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
```diff
@@ -0,0 +1,84 @@
+version: 1
```
Mohammad-Tayyab-Frequenz, line 1: Add description and add calculations for every variable.

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # ---------- Construction ----------
+    @classmethod
+    def from_yaml(  # pylint: disable=too-many-locals
+        cls,
+        path: str,
+        *,
+        locale: str = "de",
+        fallback_locale: str = "en",
+        required: Optional[Iterable[str]] = None,
```
cwasicki: `Iterable[str] | None`

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        with open(path, "r", encoding="utf-8") as f:
+            cfg = yaml.safe_load(f)
+
+        cols = cfg.get("columns") or {}
+        if not isinstance(cols, dict) or not cols:
+            raise ValueError("YAML 'columns' is missing or empty.")
+
+        c2r: Dict[str, str] = {}
+        labels_all: Dict[str, Dict[str, str]] = {}
```
cwasicki: `dict[str, str]` (here and elsewhere)

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+    """Add PV-related energy flow columns to ``df`` if PV data is present.
+
+    Derives photovoltaic (PV) energy-flow metrics from existing columns. If no PV
+    signal is present (i.e., the negative PV column is missing or all zeros), the
+    DataFrame is returned unchanged.
+
+    Args:
+      df: Input DataFrame. If present, uses columns ``pv_neg``,
+        ``consumption``, and ``battery_pos``. Missing columns are
+        treated as zeros.
```
cwasicki: I would raise if any of the expected columns is missing, otherwise it's easy to create typo bugs.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Returns:
+      The DataFrame with added PV flow columns (or unchanged if no PV signal).
+
+    Notes:
+      Newly created/updated columns:
+        - ``pv_prod``: PV production as a positive series (negated/clipped from
+          ``pv_neg``).
+        - ``pv_excess``: Excess PV after subtracting household consumption.
```
cwasicki: positive or negative?

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+-----
+These functions serve as building blocks for energy reporting, data pipelines,
+and dashboards that analyze microgrid performance, particularly in hybrid systems
+with PV, batteries, and grid interactions.
+"""
+
+import pandas as pd
+
+
+def _add_pv_energy_flows(df: pd.DataFrame) -> pd.DataFrame:
```
cwasicki: This function is problematic if there is more than one source for production, e.g. from CHP. Options to deal with this could be to
* Make this to work not with PV but "production" in general.
* Enforce that PV is the only source and otherwise raise. Then we cannot use it on certain (already existing) microgrids though, so this can only be a temporary solution.
* Take into account all sources. This requires some accounting logic though, which can be case-dependent.

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
Mohammad-Tayyab-Frequenz: Sure, will change these.

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
Mohammad-Tayyab-Frequenz: will update it

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Yeah, I picked it from the notebooks. I think previously sometimes consumption column would be missing so this logic was implemented but now we should raise if any column name is missing.
But one question would be whether pv_columns and battery columns will be present in the data even if the component is missing?

Mohammad-Tayyab-Frequenz: So, I think I need to add the logic to check whether the component type is present in the data. Then forward with the checks/calculations otherwise pass.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: So, currently the logic is to multiply pv_neg * -1 to get the pv prod in positive float.

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+  mid_consumption:
+    raw: "consumption"  # Total consumption from all the sources (grid, pv, chp, battery)
+    display:
+      en: "MID Consumption"
+      de: "MID Gesamtverbrauch"
+    description:
+      en: "Total electricity consumption from all available sources (grid, PV, CHP, battery)."
+      de: "Gesamter Stromverbrauch aus allen verfügbaren Quellen (Netz, PV, BHKW, Batterie)."
+    calculations: "grid_load + pv_self + battery_pos + chp_output (if available)."
```
Mohammad-Tayyab-Frequenz, line 36: Should I remove these calculations from all the keys? @cwasicki

cwasicki: I would remove it. If you think it's important for users to understand, it should be part of the description.

src/frequenz/lib/notebooks/reporting/utils/column_mapper.py
```diff
@@ -25,7 +25,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field, replace
-from typing import Dict, Iterable, Mapping, Optional
+from typing import Iterable, Mapping
```
cwasicki, line 28: These changes should be part of the previous commit.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,291 @@
...
+    """Return True if the production signal exists and contains any non-zero values."""
+    return isinstance(prod, pd.Series) and bool((prod != 0).any())
+
+
+def _compute_prod(signal: pd.Series, prod_is_positive: bool) -> pd.Series:
+    """Normalize any-signed production signal to positive-only production."""
+    return (signal if prod_is_positive else -signal).clip(lower=0)
+
+
+def _compute_excess(prod: pd.Series, consumption: pd.Series) -> pd.Series:
```
cwasicki, line 46: `compute` is not really adding useful information to the functions and can be removed from the names.

Also, I guess this needs more information about the excess, e.g. `prod_excess`.

cwasicki: The metric should be as clear as possible from the function name.

cwasicki: Moreover, since the idea is to use those metrics elsewhere, they should live in a dedicated metrics module and be public.

PR #175, opened 2025-10-23
Add prod metrics

Mohammad-Tayyab-Frequenz:
PR Description

- Implements production-flow metrics in [src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py](src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py), including helpers such as `add_energy_flows()` with detailed handling of missing columns and battery flows.

- Adds ColumnMapper in [src/frequenz/lib/notebooks/reporting/utils/column_mapper.py](src/frequenz/lib/notebooks/reporting/utils/column_mapper.py) plus `schema_mapping.yaml` to translate between raw API headers, canonical names, and localized display labels.


src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
```diff
@@ -0,0 +1,148 @@
...
+
+  mid_consumption:
+    raw: "consumption"  # Total consumption from all the sources (grid, pv, chp, battery)
+    display:
+      en: "MID Consumption"
+      de: "MID Gesamtverbrauch"
+    description:
+      en: "Total electricity consumption from all available sources (grid, PV, CHP, battery)."
+      de: "Gesamter Stromverbrauch aus allen verfügbaren Quellen (Netz, PV, BHKW, Batterie)."
+    implementation: "src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py::determine_consumption"
```
cwasicki: Maybe just the function name. One thing less to keep track of, e.g. if the folder structure is changed.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+      df: Input DataFrame containing the columns to be summed.
+      cols: list of column names to sum. If empty, returns a zero-filled Series.
+
+    Returns:
+      A Series representing the total sum of all specified columns.
+      If `cols` is empty, returns a Series of zeros aligned to the DataFrame index.
+    """
+    if not cols:
+        return pd.Series(0.0, index=df.index, dtype="float64")
+    series_list = [_get_series(df, c).astype("float64") for c in cols]
```
cwasicki: Or just
```
return pd.concat(series_list, axis=1).sum(axis=1)
```

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def prod_excess(prod: pd.Series, consumption: pd.Series) -> pd.Series:
+    """Compute the excess production relative to consumption.
+
+    Calculates the surplus energy or power by subtracting consumption from
+    production and removing negative results.
+
+    Args:
+      prod: Series of production values (e.g., kW or MW).
```
cwasicki: Is prod positive or negative?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+      consumption: Series of consumption values (same units as `prod`).
+
+    Returns:
+      A Series representing the excess production, where negative differences
```
cwasicki: I would always highlight the sign convention, i.e. if more excess results in a larger or smaller (if negative) value.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+    stored or exported, by subtracting excess production from total production.
+
+    Args:
+      asset_prod: Series representing total production (e.g., kW).
+      prod_excess: Series representing production surplus (portion not directly used).
+
+    Returns:
+      A Series representing self-consumed production (non-negative values only).
+    """
+    return (asset_prod - prod_excess).clip(lower=0)
```
cwasicki: Why clip lower at 0? In fact I would raise an exception or warn if there prod is smaller than excess.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+    production and removing negative results.
+
+    Args:
+      prod: Series of production values (e.g., kW or MW).
+      consumption: Series of consumption values (same units as `prod`).
+
+    Returns:
+      A Series representing the excess production, where negative differences
+      are replaced with zero.
+    """
```
cwasicki, line 67: You cannot be certain that consumption is positive.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+    Calculates the part of total production that is used locally rather than
+    stored or exported, by subtracting excess production from total production.
+
+    Args:
+      asset_prod: Series representing total production (e.g., kW).
+      prod_excess: Series representing production surplus (portion not directly used).
+
+    Returns:
+      A Series representing self-consumed production (non-negative values only).
+    """
```
cwasicki, line 154: I suggest to define self-consumption directly via prod and consumption itself.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+
+    Returns:
+      A Series representing the excess production, where negative differences
+      are replaced with zero.
+    """
+    return (prod.fillna(0) - consumption.fillna(0)).clip(lower=0)
+
+
+def prod_excess_in_bat(
+    prod_excess: pd.Series, energy_flow_into_battery: pd.Series
```
cwasicki: A better pattern would be to define the metrics on the original inputs, i.e. consumption, prod, etc. By this they are more versatile. `prod_excess` is already a metric derived from the original inputs, so you cannot calculate this metric without calculating another one first.

cwasicki: Applicable to all metrics.

Btw, you could still create the metric inside the body if you want to use it.

RELEASE_NOTES.md
```diff
@@ -9,12 +9,16 @@
...
 
 - The minimum supported version of `matplotlib` is now `v3.9.2`.
+- Add `src/frequenz/lib/notebooks/reporting/schema_mapping.yaml` to your deployment so notebooks can load the canonical column definitions via `ColumnMapper`.
 
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
 
 - This release supports python 3.13.
+- `src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py` now exposes helpers such as `add_energy_flows()` to derive production excess, grid feed-in, and self-consumption shares from raw time-series data.
```
cwasicki: First, this module is completely new, and one important addition is the definitions of all the metrics, so I would mention this. Then it also has these helper functions like `add_energy_flows`, which are more specific to the reporting notebook.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+        if len(grid_cols) != 1:
+            raise ValueError("Exactly one grid column must be specified in grid_cols.")
+
+        grid_col = grid_cols[0]
+        df["consumption"] = df[grid_col] - df[prod_cols].sum(axis=1)
+    return df
+
+
+# pylint: disable=too-many-arguments, too-many-locals
+def add_energy_flows(
```
cwasicki: Just a remark: In future I would move this to a separate module, since it's a composition of the other ones and its expected usage context is different.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,304 @@
...
+
+    If a 'consumption' column is not present, it is computed as the grid import
+    minus total production. Ensures that only expected columns exist in the
+    DataFrame and that exactly one grid column is provided.
+
+    Args:
+      df: Input DataFrame containing production, grid, and optionally 'soc' columns.
+      prod_cols: list of production column names, such as 'chp', 'pv', 'battery',
+        'ev', or any other production-related components contributing to on-site generation.
+      grid_cols: list containing exactly one grid column name.
```
cwasicki: why a list if only a single element is allowed?

src/frequenz/lib/notebooks/reporting/schema_mapping.yaml
Mohammad-Tayyab-Frequenz: Makes sense. I will change this

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: production_cols are negative in general

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Clipping at 0 ensures that negative values (which represent shortages, not excess) are set to zero. This way, the function only measures surplus production.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yes, physically it's not possible. Just wanted to make sure that there are no small errors. 
I could add a warning here.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Updated it in the latest commits.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yes, I have updated functions to work like that now.

RELEASE_NOTES.md
Mohammad-Tayyab-Frequenz: Updated the release notes.md

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Moved it to a separate utils helper folder.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Updated the function to include a list of cols.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Returns:
+        A Series where only production values (≥ 0) are retained, with all
+        non-productive values set to zero.
+    """
+    return (prod if prod_is_positive else -prod).fillna(0).clip(lower=0)
+
+
+def prod_excess(
+    prod: pd.Series, consumption: pd.Series, prod_is_positive: bool = True
```
cwasicki: Since we usually use PSC where prod is negative, I would set this default to `False`

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,235 @@
...
+        A Series representing excess production (≥ 0).
+    """
+    asset_prod_series = asset_prod(prod, prod_is_positive=prod_is_positive)
+    return (asset_prod_series - consumption.fillna(0)).clip(lower=0)
+
+
+def prod_excess_in_bat(
+    prod: pd.Series,
+    consumption: pd.Series,
+    energy_flow_into_battery: pd.Series,
```
cwasicki: This is just the battery power right? To keep naming consistent with prod and consumption, the name would be just `battery` then.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -0,0 +1,155 @@
...
+    return pd.concat(series_list, axis=1).sum(axis=1).astype("float64")
+
+
+# pylint: disable=too-many-arguments, too-many-locals
+def add_energy_flows(
+    df: pd.DataFrame,
+    production_cols: list[str] | None = None,
+    consumption_cols: list[str] | None = None,
+    battery_charge_col: str | None = None,
+    prod_is_positive: bool = True,
```
cwasicki: Isn't that default wrong, so it wouldn't work for our data?

cwasicki (commented):
LGTM, thanks a lot! Just some concern with the sign default and the naming.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yes, this is battery power and it could be different from what we get directly from the reporting api. So, would it make sense to rename the function as battery (column that we receive from reporting api is also named battery)?

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Yes, I think it should be False by default. I will update it.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Makes sense. Will change it.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
cwasicki: Actually I was referring to this argument, `energy_flow_into_battery`. Isn't this the battery power we receive from the reporting API? 

For the function name, I think the previous name was ok. This is too generic now and does not describe what it does.

cwasicki: Minor: Btw, just realized now that we are using `prod` but `consumption`. For future maybe just switch to  `production` here? But can be addressed later also.

Mohammad-Tayyab-Frequenz: Yes, makes sense. I will change that argument name.

I also noticed that discrepancy in the abbreviation for prod. I will fix it.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,243 @@
...
+            )
+
+        # Compute total grid import and total production
+        grid_total = df[grid_cols].sum(axis=1)
+        production_total = df[production_cols].sum(axis=1)
+
+        # Compute inferred consumption
+        df["consumption"] = grid_total - production_total
+
+    return df
```
cwasicki: Sorry, missed this before: This function breaks the pattern with the rest of the metrics w.r.t. it's behavior (not pure), return value (data frame) and naming. Can this just return consumption as series?

Mohammad-Tayyab-Frequenz: Yeah, I directly picked the function from the other folder, will fix it.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -0,0 +1,242 @@
...
+    production_self_use = production_self_consumption(
+        production, consumption, production_is_positive=production_is_positive
+    )
+    denom = consumption.astype("float64")
+    denom = denom.mask(denom <= 0)  # NaN when consumption <= 0
+    share = production_self_use.astype("float64") / denom
+    return share
+
+
+def determine_consumption(
```
cwasicki: You can remove the `determine_` from the metric.

cwasicki (dismissed):
LGTM

cwasicki (approved):

PR #176, opened 2025-10-28
Switch to schema-based microgrid config loading

cwasicki:
Previously, microgrid configs were initialized by unpacking nested dictionaries. Now the class defines a marshmallow schema and uses it to deserialize toml data into valudated python objects.

Introduces a new `load_from_file` function to load multiple configs as dict from a single file. This is also used now in the existing function to load multiple files or directories.


Mohammad-Tayyab-Frequenz (approved):
LGTM

PR #177, opened 2025-10-29
fix: missing production and consumption column

Mohammad-Tayyab-Frequenz:


cwasicki:
This needs a rebase

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -41,16 +45,25 @@ def _get_numeric_series(df: pd.DataFrame, col: str | None) -> pd.Series:
...
+        series = df.reindex(columns=[col], fill_value=0)[col].astype("float64")
+
+    if clip_non_negative:
+        series = series.clip(lower=0)
+    return series
 
 
-def _sum_cols(df: pd.DataFrame, cols: list[str] | None) -> pd.Series:
+def _sum_cols(
+    df: pd.DataFrame, cols: list[str] | None, *, clip_non_negative: bool = False
```
cwasicki: The new argument is not used anywhere, so why introducing it?

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -41,16 +45,25 @@ def _get_numeric_series(df: pd.DataFrame, col: str | None) -> pd.Series:
...
     """
     if col is None:
-        return pd.Series(0.0, index=df.index, dtype="float64")
-    return df.reindex(columns=[col], fill_value=0)[col].astype("float64").clip(lower=0)
+        series = pd.Series(0.0, index=df.index, dtype="float64")
+    else:
+        series = df.reindex(columns=[col], fill_value=0)[col].astype("float64")
+
+    if clip_non_negative:
+        series = series.clip(lower=0)
```
cwasicki: IIUC you only use it in one place, and I also don't see the point of having this flag (which is just a boolean for lower=0) when you could simply do `_get_numeric_series(...).clip(lower=0)`, which is more flexible and readable.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -106,15 +126,71 @@ def add_energy_flows(
...
     df_flows = df.copy()
 
-    # Total production and consumption (returns pandas series with 0.0 for missing cols)
-    df_flows["production_total"] = _sum_cols(df_flows, production_cols)
-    df_flows["consumption_total"] = _sum_cols(df_flows, consumption_cols)
-
-    # Surplus vs. consumption
+    # Normalize production, grid and consumption columns by removing None entries
+    resolved_production_cols = [
+        col for col in (production_cols or []) if col is not None
```
cwasicki: Why are there `None`s in the columns?

Mohammad-Tayyab-Frequenz: This is just for handling the cases where users do not give any production/consumption/grid cols.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Yes, makes sense. I will do that

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -96,7 +96,11 @@ def production_excess_in_bat(
     production_excess_series = production_excess(
         production, consumption, production_is_positive=production_is_positive
     )
-    battery = battery.astype("float64").clip(lower=0)
+    battery = (
+        battery.astype("float64")
+        .reindex(production_excess_series.index, fill_value=0.0)
+        .clip(lower=0)
+    )
     return pd.concat([production_excess_series, battery], axis=1).min(axis=1)
```
Mohammad-Tayyab-Frequenz, line 100: Added this to ensure that battery is properly aligned with production_excess_series.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -199,44 +203,60 @@ def production_self_share(
 
 
 def consumption(
```
Mohammad-Tayyab-Frequenz, line 201: Added battery series separately and changed the function to work on pd.series rather than pd.dataframes.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -106,47 +123,83 @@ def add_energy_flows(
...
+        consumption_series_cols.append(col)
+
+    df_flows["consumption_total"] = _sum_cols(df_flows, consumption_series_cols)
+
+    # Surplus vs. consumption (production is already positive because of the above cleaning)
     df_flows["production_excess"] = production_excess(
         df_flows["production_total"],
         df_flows["consumption_total"],
-        production_is_positive=production_is_positive,
+        production_is_positive=True,
```
Mohammad-Tayyab-Frequenz, line 171: Initial input production is negative but is handled by asset_production to invert the sign. So, we have to change the arguments to positive.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -106,47 +123,83 @@ def add_energy_flows(
...
     )
 
     # Battery charging power (optional)
-    bat_in = _get_numeric_series(df_flows, battery_charge_col)
```
Mohammad-Tayyab-Frequenz, line 121: Created a separate battery series containing only positive values to represent the battery charge.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
cwasicki: Why can they be not aligned?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -199,44 +203,60 @@ def production_self_share(
...
-    empty production columns by treating them as zero.
+    Computes: consumption = grid_import - production (raw production-neg values)
+                            - battery (raw - with positive and negative values).
 
     Args:
-        df: Input DataFrame containing grid and optional production columns.
-        production_cols: List of production column names (e.g., "pv", "chp", "battery" or "ev").
-            Can be None or empty if no on-site generation is present.
-        grid_cols: List of one or more grid column names.
+        grid: Series of grid import values (e.g., kW or MW).
```
cwasicki: Why only grid import and not grid power?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -199,44 +203,60 @@ def production_self_share(
...
+            which may indicate times of net export or a sign-convention mismatch.
     """
-    if "consumption" in df.columns:
-        return df["consumption"]
+    if grid is None:
+        raise ValueError("`grid` must be provided as a pandas Series.")
 
-    if not grid_cols:
-        raise ValueError("At least one grid column must be specified in grid_cols.")
+    grid_s = grid.astype("float64").fillna(0)
```
cwasicki: I think we should not fill with zeroes. There can be various reasons that the grid measurements are missing and we cannot safely know if they were zero.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -199,44 +203,60 @@ def production_self_share(
...
-    # Compute inferred consumption (Series)
-    consumption = grid_total - production_total
-    consumption.name = "consumption"
+    result = (grid_s - prod_s - battery_s).astype("float64")
+    result.name = "consumption"
 
-    return consumption
+    if (result < 0).any():
+        warnings.warn(
+            "Negative inferred consumption detected. This can occur during net export "
```
cwasicki: What do you mean by net export and why can this cause negatives here? I think the most likely case to get negative values here is unmeasured energy generation sources or short-term spikes.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -43,11 +44,14 @@ def _get_numeric_series(df: pd.DataFrame, col: str | None) -> pd.Series:
...
-        A float64 Series with non-negative values, matching the input index.
+        A float64 Series aligned to the input index.
     """
     if col is None:
-        return pd.Series(0.0, index=df.index, dtype="float64")
-    return df.reindex(columns=[col], fill_value=0)[col].astype("float64").clip(lower=0)
+        series = pd.Series(0.0, index=df.index, dtype="float64")
+    else:
+        raw = df.reindex(columns=[col], fill_value=0)[col]
+        series = pd.to_numeric(raw, errors="coerce").fillna(0.0).astype("float64")
```
cwasicki, line 53: Not introduced in this PR, but also here not sure if we should by default fill with 0s.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -67,32 +71,45 @@ def _sum_cols(df: pd.DataFrame, cols: list[str] | None) -> pd.Series:
...
         return pd.Series(0.0, index=df.index, dtype="float64")
 
-    # Safely extract each column as a numeric, non-negative Series, then sum row-wise
+    # Safely extract each column as a numeric Series then sum row-wise
     series_list = [_get_numeric_series(df, c) for c in cols]
     return pd.concat(series_list, axis=1).sum(axis=1).astype("float64")
 
 
-# pylint: disable=too-many-arguments, too-many-locals
+def _column_has_data(df: pd.DataFrame, col: str | None) -> bool:
```
cwasicki, line 79: Maybe `_has_nonzero_values(df, *, column: str | None)`? And why would you call it with None at all?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yeah, I wasn’t sure if reindexing would be the best approach moving forward or not.
Consider cases where users provide input data in which production and consumption have matching indices and are aligned correctly, but the battery data contains extra rows (for any unknown reason). In such cases, the downstream functions would end up with more rows than the original production and consumption data.

Should we handle such scenarios within this Python codebase, or can we assume that the input data is already clean and pre-processed?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yeah, it should be grid power. I will update the docstring.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: We need to fill it with 0 so as to avoid the cases of the consumption also being NaN in those cases.
For eg. 
Production - 15, battery - 5, grid - NaN
Then consumption will also be NaN.

So, in this case should we have the consumption as NaN or 20?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: I think the most likely case of negative consumption is only going to be when the production sign was positive and after subtracting it from grid we get a negative value.

Net export are the cases where there was less consumption than production on-site and hence, the consumption had a negative sign. I just added this as a warning to make sure that users know that they have such periods and to check if there was no calculation error.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: This is also to make sure we don't get NaNs during downstream calculations.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: This could occur in cases where the any column input (production, battery, consumption) from the user was not provided and by default it is assumed as None. 
So, we need to safely pass those cases from our function

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
cwasicki: Consumption should be NaN, because you don't know what the grid power is.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
cwasicki: > and hence, the consumption had a negative sign.

Do you mean "grid power had a negative sign"?

And when you say net export, does it refer to the negative part of grid power?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
cwasicki: Which values would the rows have? If it's NaNs I wouldn't be concerned.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yeah, that makes sense. I'm not sure, though, if the NaN values in the consumption column might cause errors in the downstream functions.

In the overview section, where we display the total consumption, should we ignore these values and show the rest of the data, or should we display a warning instead?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yes, exactly.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
cwasicki: Why not accepting `NaN`s if something goes wrong? Or do we know that `NaN` should be 0? 

Also, why do we need `to_numeric`? IMO we should expect only numeric input data and can fail otherwise.

cwasicki (approved):

PR #190, opened 2025-11-12
Add grid import function

Mohammad-Tayyab-Frequenz:
Added grid import function in the reporting metrics to calculate the amount of electricity purchased from the grid. 

This would be used to calculate grid_import from the given production, consumption and battery columns.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,12 +35,13 @@ dependencies = [
   "pvlib >= 0.13.0, < 0.14.0",
   "python-dotenv >= 0.21.0, < 1.3.0",
   "toml >= 0.10.2, < 0.11.0",
-  "marshmallow_dataclass >= 8.7.1, < 9",
+  "marshmallow-dataclass>=8.7.1,<9",
   "plotly >= 6.0.0, < 6.4.0",
   "kaleido >= 0.2.1, < 1.2.0",
   "frequenz-client-reporting >= 0.19.0, < 0.20.0",
   "frequenz-client-weather >= 0.2.3, < 0.3.0",
   "types-pyyaml>=6.0.12.20250915",
+  "marshmallow>=4.1.0,<5",
 ]
 dynamic = ["version"]
 
--- a/src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
+++ b/src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
@@ -244,3 +244,54 @@ def consumption(
     result.name = "consumption"
 
     return result
+
+
+def grid_consumption(
+    grid: pd.Series | None,
+    production: pd.Series | None,
+    consumption: pd.Series | None,
+    battery: pd.Series | None,
+) -> pd.Series:
+    """Determine grid import using measured or derived series.
+
+    This function follows the PSC (Production-Storage-Consumption) convention:
+    - Production is positive when generating.
+    - Battery is positive when charging (consuming energy).
+    - Consumption is positive when consuming energy.
+
+    Prefers the measured grid series (positive import, negative export). When
+    unavailable, derives grid import from the remaining energy balance.
+
+    Args:
+        grid: Grid power series (positive import), or ``None``.
+        production: Production power series (PSC-convention), or ``None``.
+        consumption: Consumption power series (PSC-convention), or ``None``.
+        battery: Battery power (PSC-convention: charging positive), or ``None``.
+
+    Returns:
+        Series with non-negative grid import values.
+
+    Raises:
+        ValueError: If neither a grid series nor any of production,
+            consumption, or battery series are provided.
+    """
+    if grid is not None:
+        return grid.astype("float64").clip(lower=0)
+
+    # Ensure at least one other input is provided
+    if all(s is None for s in (production, consumption, battery)):
+        raise ValueError(
+            "Cannot compute grid_consumption because no grid, production, "
+            "consumption, or battery series were provided. "
+            "This function follows PSC conventions, so production and battery "
+            "must also follow those conventions."
+        )
+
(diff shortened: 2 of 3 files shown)


src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -244,3 +244,48 @@ def consumption(
...
+            UserWarning,
+            stacklevel=2,
+        )
+        return pd.Series(dtype="float64")
+
+    prod = production.astype("float64") if production is not None else 0.0
+    cons = consumption.astype("float64") if consumption is not None else 0.0
+    batt = battery.astype("float64") if battery is not None else 0.0
+
+    inferred = cons - prod - batt
```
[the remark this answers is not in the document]
Mohammad-Tayyab-Frequenz: @cwasicki Do you think it should be addition and not subtraction?

pyproject.toml
```diff
@@ -35,12 +35,13 @@ dependencies = [
...
   "python-dotenv >= 0.21.0, < 1.3.0",
   "toml >= 0.10.2, < 0.11.0",
-  "marshmallow_dataclass >= 8.7.1, < 9",
+  "marshmallow-dataclass>=8.7.1,<9",
   "plotly >= 6.0.0, < 6.4.0",
   "kaleido >= 0.2.1, < 1.2.0",
   "frequenz-client-reporting >= 0.19.0, < 0.20.0",
   "frequenz-client-weather >= 0.2.3, < 0.3.0",
   "types-pyyaml>=6.0.12.20250915",
+  "marshmallow>=4.1.0",
```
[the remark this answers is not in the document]
cwasicki: Agree

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -244,3 +244,48 @@ def consumption(
...
+    battery: pd.Series | None,
+) -> pd.Series:
+    """Determine grid import using measured or derived series.
+
+    Prefers the measured grid series (positive import, negative export). When
+    unavailable, derives grid import from the remaining energy balance.
+
+    Args:
+        grid: Grid power series (positive import), or ``None``.
+        production: Production power series, or ``None``.
```
cwasicki: Is this according to Passive Sign Convention or is production here positive?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -244,3 +244,48 @@ def consumption(
...
+
+    Warns:
+        UserWarning: If neither a grid series nor any of production,
+            consumption, or battery series are provided.
+    """
+    if grid is not None:
+        return grid.astype("float64").clip(lower=0)
+
+    # Ensure at least one other input is provided
+    if all(s is None for s in (production, consumption, battery)):
```
cwasicki, line 282: This basically means you call this function only with `None`s, IMO we should raise in this case.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
```diff
@@ -244,3 +244,48 @@ def consumption(
     result.name = "consumption"
 
     return result
+
+
+def grid_import(
```
cwasicki: In other places we refer to this as grid consumption. This term also makes sense, but ideally we keep it consistent. Is this term used more widely?

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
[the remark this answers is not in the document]
cwasicki: Depends on the sign of production. If the inputs all follow PSC (which I strongly recommend), it should be `cons+prod+batt`. If prod is non-negative (inverted PSC), copilot suggestion is correct.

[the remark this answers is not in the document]
Mohammad-Tayyab-Frequenz: I was calling the function after converting production to positive. But it makes sense to keep the reporting metrics as generic and standard as possible. So, I will change it to PSC.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Updated the docstring and code with PSC.

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Fixed

src/frequenz/lib/notebooks/reporting/metrics/reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yeah, I am not sure. Previously it was also called net_import in the reporting notebooks.

Grid consumption is also just the positive part of the grid load, right? (excluding whatever was fed into the grid)

cwasicki: Yes

pyproject.toml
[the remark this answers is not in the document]
cwasicki: This still needs fixing

cwasicki (commented):
LGTM, just the dependency range fix.

pyproject.toml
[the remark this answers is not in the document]
Mohammad-Tayyab-Frequenz: I think this got overwritten with a rebase. Fixed it now.

cwasicki (approved):

PR #198, opened 2025-11-27
tests: add unit test cases for reporting metrics

Mohammad-Tayyab-Frequenz:
This PR adds comprehensive unit test coverage for the reporting metrics module, which provides energy flow calculations for hybrid power systems including production, consumption, battery storage, and grid interactions.

- Introduces 9 test cases covering all key functions in the reporting_metrics module
- Tests edge cases including sign conventions, missing values, clipping behavior, and error conditions
- Validates warning emissions and error handling for invalid inputs

### DIFF
--- a/tests/test_reporting_metrics.py
+++ b/tests/test_reporting_metrics.py
@@ -1,0 +1,115 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the reporting metrics helpers."""
+
+from __future__ import annotations
+
+import pandas as pd
+import pytest
+from pandas.testing import assert_series_equal
+
+from frequenz.lib.notebooks.reporting.metrics import reporting_metrics as metrics
+
+
+def test_asset_production_respects_sign_and_missing_values() -> None:
+    """Positive flag controls sign inversion and NaNs stay missing."""
+    production = pd.Series([10, -5, None], index=pd.RangeIndex(3))
+
+    positive = metrics.asset_production(production, production_is_positive=True)
+    expected_positive = pd.Series([10.0, 0.0, float("nan")], index=production.index)
+    assert_series_equal(positive, expected_positive)
+
+    negative = metrics.asset_production(production, production_is_positive=False)
+    expected_negative = pd.Series([0.0, 5.0, float("nan")], index=production.index)
+    assert_series_equal(negative, expected_negative)
+
+
+def test_production_excess_clips_negative_surplus() -> None:
+    """Excess is computed after sign correction and clipped at zero."""
+    production = pd.Series([-5.0, -1.0, -10.0], index=pd.RangeIndex(3))
+    consumption = pd.Series([3.0, 2.0, 1.0], index=production.index)
+
+    result = metrics.production_excess(production, consumption)
+    expected = pd.Series([2.0, 0.0, 9.0], index=production.index)
+    assert_series_equal(result, expected)
+
+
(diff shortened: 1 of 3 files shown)


tests/test_reporting_metrics.py
```diff
@@ -0,0 +1,115 @@
...
+
+import pandas as pd
+import pytest
+from pandas.testing import assert_series_equal
+
+from frequenz.lib.notebooks.reporting.metrics import reporting_metrics as metrics
+
+
+def test_asset_production_respects_sign_and_missing_values() -> None:
+    """Positive flag controls sign inversion while NaNs become zeros."""
```
cwasicki: How do we know that we can turn NaNs into zeroes?

tests/test_reporting_metrics.py
```diff
@@ -0,0 +1,115 @@
...
+    production = pd.Series([10, -5, None], index=pd.RangeIndex(3))
```
cwasicki, line 17: Are `None`s turned into `NaN`s?

tests/test_reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yes, I remember now we discussed that we should not convert to zero. 
I will fix it in the code.

tests/test_reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yes, but with the previous comment the test case would change.

tests/test_reporting_metrics.py
```diff
@@ -0,0 +1,115 @@
...
+
+def test_asset_production_respects_sign_and_missing_values() -> None:
+    """Positive flag controls sign inversion and NaNs stay missing."""
+    production = pd.Series([10, -5, None], index=pd.RangeIndex(3))
+
+    positive = metrics.asset_production(production, production_is_positive=True)
+    expected_positive = pd.Series([10.0, 0.0, float("nan")], index=production.index)
+    assert_series_equal(positive, expected_positive)
+
+    negative = metrics.asset_production(production, production_is_positive=False)
```
cwasicki, line 23: Not for this PR, but I feel we should remove the `production_is_positive` flag and stick to PSC wherever possible. And in the other cases we can just explicitly reverse the sign, correct?

tests/test_reporting_metrics.py
```diff
@@ -0,0 +1,115 @@
...
+    assert_series_equal(result, expected)
+
+
+def test_grid_feed_in_excludes_energy_stored_in_battery() -> None:
+    """Grid feed-in equals production surplus minus what is stored in the battery."""
+    production = pd.Series([-8, -3], index=pd.RangeIndex(2))
+    consumption = pd.Series([2, 1], index=production.index)
+    battery = pd.Series([5, 10], index=production.index)
+
+    result = metrics.grid_feed_in(production, consumption, battery)
```
cwasicki, line 55: In future we might generalize this to storage instead of battery only.

cwasicki (approved):
LGTM, thank you!

tests/test_reporting_metrics.py
Mohammad-Tayyab-Frequenz: Yes, we can do that.

tests/test_reporting_metrics.py
Mohammad-Tayyab-Frequenz: Got it. I will make note of it as well.

PR #228, opened 2025-12-22
Fill missing aggregated values

Mohammad-Tayyab-Frequenz:
This pull request adds functionality to fill missing values in aggregated component columns by summing their corresponding individual component data. The implementation automatically backfills NaN values in aggregate columns (e.g., battery_power_flow, pv_asset_production) by computing sums from labeled component columns (e.g., Battery #1, PV #1).

Key changes:

- Introduces fill_aggregated_component_columns() helper function with configurable component type filtering
- Integrates the filling logic into the energy report DataFrame creation pipeline with an optional flag
- Adds comprehensive test coverage for both basic filling behavior and component type filtering

### DIFF
--- a/src/frequenz/lib/notebooks/reporting/utils/helpers.py
+++ b/src/frequenz/lib/notebooks/reporting/utils/helpers.py
@@ -39,7 +39,7 @@ from __future__ import annotations
 
 import warnings
 from datetime import date, datetime, time
-from typing import Any, Literal, cast
+from typing import Any, Literal, Mapping, cast
 
 import matplotlib.colors as mcolors
 import pandas as pd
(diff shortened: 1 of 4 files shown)


src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -639,3 +639,47 @@ def to_rgba_str(color: str) -> str:
...
+            (e.g., "battery_power_flow") and labeled individual component columns
+            (e.g., "Battery #1", "Battery #2").
+        component_types: List of component types to consider for aggregation
+            (e.g., ["battery", "pv"]).
+
+    Returns:
+        DataFrame with missing aggregated component columns filled in by summing
+        the corresponding individual component columns.
+    """
+    # Mapping configuration
```
cwasicki: This mapping is probably not specific to only this function, so should be passed in as an argument.

Mohammad-Tayyab-Frequenz: Fixed it in the latest commits.

cwasicki (approved):

PR #261, opened 2026-02-24
Normalize report date handling and update color map docs

Mohammad-Tayyab-Frequenz:
Updates reporting utilities to improve date/time handling and plotting defaults with centralized color configuration.

Changes:

- Replaced pytz usage with zoneinfo and added `normalize_date_for_report()` to return current time for today, midnight for past dates, and reject future dates.
- Update `build_color_map()` with default color_dict.
- Extended `plot_time_series()` with dotted_cols and tweaked fill behavior / layout positioning.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,7 +11,8 @@
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
+- `build_color_map` now sources default colors from the in-code color dictionary rather than a YAML file.
 
 ## Bug Fixes
 
-Fix state analysis to support new `ElectricalComponent*Code` enums and diagnostic code handling.
+- Added `normalize_date_for_report` to return current time for today, midnight for past dates, and reject future dates.
--- a/src/frequenz/lib/notebooks/reporting/plotter.py
+++ b/src/frequenz/lib/notebooks/reporting/plotter.py
@@ -165,7 +165,7 @@ def plot_time_series(
                 bgcolor="rgba(0,0,0,0)",  # Transparent background
                 activecolor="#2C7BE5",  # Highlight color for active button
                 font=dict(size=12),
-                x=0.65,
+                x=0,
                 xanchor="left",
                 y=1.05,
                 yanchor="top",
--- a/src/frequenz/lib/notebooks/reporting/plotter.py
+++ b/src/frequenz/lib/notebooks/reporting/plotter.py
@@ -6,8 +6,23 @@
 import pandas as pd
 import plotly.express as px
 import plotly.graph_objects as go
+from matplotlib import colors as mcolors
 
 from frequenz.lib.notebooks.reporting.utils.helpers import build_color_map, long_to_wide
+
+
+def _with_alpha(color: str | None, alpha: float) -> str | None:
+    """Return color as rgba string with the given alpha, or None if invalid."""
+    if not color:
+        return None
+    try:
+        r, g, b, _ = mcolors.to_rgba(color)
+    except ValueError:
+        return None
+    r255 = int(round(r * 255))
+    g255 = int(round(g * 255))
+    b255 = int(round(b * 255))
+    return f"rgba({r255},{g255},{b255},{alpha:.3f})"
 
 
 # pylint: disable=too-many-arguments, too-many-positional-arguments,
@@ -98,11 +113,9 @@ def plot_time_series(
 
     # Add one line trace per column
     for i, col in enumerate(cols):
-        fill_mode = "tonextx" if col in fill_cols else "none"
+        fill_mode = "tozeroy" if col in fill_cols else "none"
         line_color = color_map.get(col)
-        fill_color = (
-            line_color.replace("1)", "0.3)") if isinstance(line_color, str) else None
-        )
+        fill_color = _with_alpha(line_color, 0.9)
 
         fig.add_trace(
             go.Scatter(
--- a/src/frequenz/lib/notebooks/reporting/plotter.py
+++ b/src/frequenz/lib/notebooks/reporting/plotter.py
@@ -40,6 +40,7 @@ def plot_time_series(
     category_col: str | None = None,
     value_col: str | None = None,
     fill_cols: list[str] | None = None,
+    dotted_cols: list[str] | None = None,
     plot_order: list[str] | None = None,
 ) -> go.Figure:
     """Create an interactive time-series plot using Plotly.
@@ -68,6 +69,8 @@ def plot_time_series(
             format. Used only if `long_format_flag=True`.
         fill_cols: List of column names to plot as filled areas under the curve.
             Defaults to None (no fill).
+        dotted_cols: List of column names to render with dotted lines.
+            Defaults to None (no dotted lines).
         plot_order: Optional list specifying the order of columns to plot. If None,
             the order in `cols` is used.
 
@@ -110,6 +113,9 @@ def plot_time_series(
     # Check if fill_cols is provided
     if fill_cols is None:
         fill_cols = []
+    if dotted_cols is None:
+        dotted_cols = []
+    dotted_set = set(dotted_cols)
 
     # Add one line trace per column
     for i, col in enumerate(cols):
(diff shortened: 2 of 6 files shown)


src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -511,21 +512,27 @@ def set_date_to_midnight(
...
             RuntimeWarning,
         )
-        tz = pytz.UTC
+        tz = ZoneInfo("UTC")
 
-    return tz.localize(datetime.combine(input_date, time.min))
+    now = datetime.now(tz)
+
+    if input_date == now.date():
+        return now
```
cyiallou, line 572: Why are you returning `now` in this case and not midnight as expected by this function? What is the "fix" here?

Also, if you need this option perhaps it's better to split it to a new helper and pass `now` as an input argument to make testing easier.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -39,11 +39,11 @@
 import warnings
 from datetime import date, datetime, time
 from typing import Any, Literal, Mapping, cast
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 import matplotlib.colors as mcolors
 import pandas as pd
 import plotly.express as px
-import pytz
```
cyiallou, line 46: Replacing `pytz` with `zoneinfo` should have been in a separate commit as it's not part of the fix.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -546,6 +548,19 @@ def set_date_to_midnight(
     "var",
 ]
 
+_COLOR_MAPPING_PATH = Path(__file__).with_name("colors.yaml")
+
+
+@lru_cache
```
cyiallou: Since you are caching it here, please check what happens if the YAML file was missing. I think that if the file was missing once, the funciton will keep returning `{}` for the lifetime of the process (unless you call `_load_color_dict.cache_clear()`), even if `colors.yaml` appears later.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -546,6 +548,19 @@ def set_date_to_midnight(
...
+def _load_color_dict() -> dict[str, str]:
+    """Load the default color mapping from YAML, if available."""
+    try:
+        data = load_config(str(_COLOR_MAPPING_PATH))
+    except FileNotFoundError:
```
cyiallou: Is it intended to fail completely when `load_config` raises a `TypeError` or any other unexpected error? Can't we log it as warning (and inform the user) and continue with plotting using the default palette? This is based on the assumption that `build_color_map` is used downstream and not `_load_color_dict` directly.

src/frequenz/lib/notebooks/reporting/plotter.py
cyiallou, line 30: Just a quick thought (not a formal review):

The function works and is clearly feature-rich, but it currently bundles several responsibilities together (data preparation, reshaping, column selection/order, color handling, trace creation, and layout configuration). That makes it a bit harder to read, test, and evolve over time.

A small improvement that might help would be to split parts of the logic into focused helpers (e.g., data preparation, column resolution, trace construction, layout setup). This would likely reduce cognitive load and make future changes easier without altering behavior.

src/frequenz/lib/notebooks/reporting/plotter.py
```diff
@@ -129,6 +129,7 @@ def plot_time_series(
                 y=pdf[col],
                 mode="lines",
                 name=col,
+                hovertemplate=f"<b>{col}</b>: %{{y}} {yaxis_title}<extra></extra>",
```
cyiallou, line 132: Quick question, does this actually render properly on Deepnote?

cyiallou (commented):
I have a couple of comments and a thought. Most importantly, it would help me if you can explain the motivation behind the change in `set_date_to_midnight`.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: This function is updated now to align with the reporting api changes. 
1. For all the dates prior to current date, we need the complete day's data but future times are rejected in the current reporting api. So, we can only give the end timestamp till the current time.

src/frequenz/lib/notebooks/reporting/plotter.py
Mohammad-Tayyab-Frequenz: Yes, the hover template looks like this:
image

src/frequenz/lib/notebooks/reporting/plotter.py
Mohammad-Tayyab-Frequenz: Yes, refactoring this is definitely necessary and is in my TODO tasks. I would try to include this in the coming PRs

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Yeah, I think we can get-by with a warning in this case as it makes sense.

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: _lru_cache_ was suggested by Copilot review here. I can remove this

src/frequenz/lib/notebooks/reporting/utils/helpers.py
cyiallou: How and where is this function used?

src/frequenz/lib/notebooks/reporting/utils/helpers.py
cyiallou: It makes sense to cache it, I am just saying that you probably need to clear the cache if `colors.yaml` gets updated on the fly (if possible)

src/frequenz/lib/notebooks/reporting/plotter.py
cyiallou: Cool. Deepnote must have fixed this because it was not possible a few months back

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: After we get the date input from users in deepnote/streamlit. In the below function:

```
df = await mdata.ac_active_power(
    microgrid_id=microgrid_id, 
    component_types=component_types,
    start=start_date,
    end=end_date + timedelta(days=1), # To include the data from the end date
    resampling_period=processed_resolution,
    keep_components=True,
    splits=True,
)
```

src/frequenz/lib/notebooks/reporting/utils/helpers.py
Mohammad-Tayyab-Frequenz: Done

src/frequenz/lib/notebooks/reporting/utils/helpers.py
```diff
@@ -642,13 +643,13 @@ def build_color_map(
     """Generate a color mapping for columns or categories.
 
     Creates a mapping from column names (or categorical labels) to color
-    values. If user-specified colors are provided via `color_dict`, those
-    are applied first. Remaining columns are assigned distinct colors from
-    a chosen palette, ensuring no duplicates.
+    values. Default colors are loaded from a YAML color mapping file and
```
cyiallou: This needs to be changed since we are not loading from a YAML file anymore.

RELEASE_NOTES.md
```diff
@@ -11,7 +11,8 @@
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
+- Added YAML-based color mappings (`colors.yaml`) for reporting plots, used by `build_color_map`.
```
cyiallou: Also needs to change

RELEASE_NOTES.md
```diff
@@ -11,7 +11,8 @@
...
 
 ## Bug Fixes
 
-Fix state analysis to support new `ElectricalComponent*Code` enums and diagnostic code handling.
+- `set_date_to_midnight` now returns the current time when the input date is today, otherwise midnight.
```
cyiallou: this one too

cyiallou (changes requested):
A couple of minor changes to adjust the language after the latest changes. Please also update the PR description.

cyiallou (approved):

PR #280, opened 2026-03-19
Add load config formula

Mohammad-Tayyab-Frequenz:


cwasicki (dismissed):
Cool, thanks a lot!

cwasicki (dismissed):

cwasicki (dismissed):

cwasicki:
Resolved conflicts and reverted the moving of the module. Will addresss this separately.

cwasicki (dismissed):

cwasicki (dismissed):

src/frequenz/lib/notebooks/reporting/asset_optimization/data.py
```diff
@@ -19,12 +19,14 @@
 
 async def init_microgrid_data(
     *,
-    microgrid_config_dir: str,
+    microgrid_config_file: str | None = None,
+    microgrid_config_dir: str | None = None,
```
cyiallou, line 23: Is there a use case where we prefer to pass a directory instead of a list of file paths? Do we want to allow both to be defined by the user or should we allow only one of the two? If I understood the method correctly [here](https://github.com/frequenz-floss/frequenz-gridpool-python/blob/v0.x.x/src/frequenz/gridpool/_microgrid_config.py#L348) passing a file with MIDs that are also in a file inside the specified dir will silently replace the former with the contents found in the latter.

Also, `load_configs` already allows `microgrid_config_file` to be a `list[str]`, why are we restricting to only `str` here?

cyiallou (commented):
I have a couple of questions that might touch the implementation in the gridpool repo

src/frequenz/lib/notebooks/reporting/asset_optimization/data.py
cwasicki: > Is there a use case where we prefer to pass a directory instead of a list of file paths?

One advantage I see is that it saves you the file listing before calling it. But I also have mixed feelings about the interface and am happy to support files only (maybe check in the PR where it has been added https://github.com/frequenz-floss/frequenz-lib-notebooks/pull/104/changes).

In fact I am planning to change the interface subsequently for this new function. Atm there is code which uses the folder argument, so I would leave it here. In future I only want to support a single file and remove folder support. Not a list of files, because I don't think it's necessary. Note that we create a data instance for a single microgrid, so not sure why we should support multiple files. But happy to change if you think we should.

cyiallou (dismissed):

cyiallou (approved):

PR #305, opened 2026-05-08
Add data fetching wrapper

Mohammad-Tayyab-Frequenz:
The `metric_data` function has been extended from a direct power-only fetcher into a more flexible entry point that can now work in two modes: fetching active power directly or deriving power from cumulative energy data. With the new data_fetch_mode parameter, we can either keep the existing `AC_POWER_ACTIVE` path or fetch `AC_ENERGY_ACTIVE_CONSUMED` and `AC_ENERGY_ACTIVE_DELIVERED` in parallel.

The original functionality of metric_data is shifted to `_fetch_metric_data` function now.


src/frequenz/data/microgrid/component_data.py
```diff
@@ -240,3 +242,200 @@ async def soc(  # pylint: disable=too-many-arguments
...
+            end=end,
+            component_types=component_types,
+            resampling_period=resampling_period,
+            metric="AC_ENERGY_ACTIVE_CONSUMED",
+            keep_components=keep_components,
+            splits=splits,
+        )
+        if df is None:
+            return df
+        if unit == "Wh":
```
cwasicki: This code is duplicate and should be in a function.

src/frequenz/data/microgrid/component_data.py
```diff
@@ -212,7 +214,7 @@ async def ac_active_power(  # pylint: disable=too-many-arguments
 
         if unit == "W":
```
cwasicki: Probably better to move this into a separate function.

src/frequenz/data/microgrid/component_data.py
```diff
@@ -240,3 +242,200 @@ async def soc(  # pylint: disable=too-many-arguments
...
+            else:
+                df = energy_consumed.sub(energy_delivered, fill_value=0)
+        else:
+            raise ValueError(f"Unknown metric: {metric}")
+
+        if df is None:
+            return None, [], component_types
+
+        present = [ctype for ctype in component_types if ctype in df.columns]
+        missing = [ctype for ctype in component_types if ctype not in df.columns]
```
cwasicki: What would be the reason to have component types missing?

src/frequenz/data/microgrid/component_data.py
```diff
@@ -240,3 +242,200 @@ async def soc(  # pylint: disable=too-many-arguments
             keep_components=keep_components,
         )
         return df
+
+    async def ac_active_energy_consumed(  # pylint: disable=too-many-arguments
```
cwasicki, line 357: Are you planning to use these in the reporting notebook? Is the idea to migrate everything from power to energy? In fact I thought we provide an additional possibility to get ac active power from energy. That way we can leave all downstream usage the same but only switch the data fetching.

src/frequenz/data/microgrid/component_data.py
Mohammad-Tayyab-Frequenz: yeah, makes sense.

src/frequenz/data/microgrid/component_data.py
Mohammad-Tayyab-Frequenz: I am not sure of the exact reasons but there were cases in the past where component IDs were present but the corresponding data was missing.

src/frequenz/data/microgrid/component_data.py
Mohammad-Tayyab-Frequenz: With the `metric_with_present_component_types` wrapper, we can switch between power and energy metrics whenever needed.

One thing that may currently be missing is the conversion from energy back to power after fetching the data.

This means we can fetch data using either power or energy metrics, while the wrapper consistently returns the converted power value (kW), so downstream functions do not need to be modified.

cwasicki: Yea, the conversion is indeed missing, probably with a strategy to fill missing values. To me the method is also conflating too many things: 1) Do we have a use-case for other metrics than ac active power? 2) We should understand why components are missing and not hide it with this function.

Mohammad-Tayyab-Frequenz: I have added the conversion function but no method/strategy for filling missing values are added yet. I feel we should discuss that before implementation.

For point 1:
1. I think in future we might want to move towards `ac_active_energy_delivered` for feed-in and `ac_active_energy_consumed` for grid consumption since within the 15 mins cycle we may buy and sell as seen in the Agravis data. 
2. I have removed the logic for missing components and will test it out to see if we still get error for any MID.

src/frequenz/data/microgrid/component_data.py
Mohammad-Tayyab-Frequenz: I have removed the removal of missing component logic now.

Mohammad-Tayyab-Frequenz:
@cwasicki Can you please have a look at this PR again?
Thank you!

src/frequenz/data/microgrid/component_data.py
```diff
@@ -52,35 +60,42 @@ def microgrid_configs(self) -> dict[str, MicrogridConfig]:
...
+        cols = df.columns
+        pos_cols = [f"{col}_pos" for col in cols]
+        neg_cols = [f"{col}_neg" for col in cols]
+        df[pos_cols] = df[cols].clip(lower=0)
+        df[neg_cols] = df[cols].clip(upper=0)
+        return df
+
     # pylint: disable=too-many-locals
-    async def metric_data(  # pylint: disable=too-many-arguments
+    async def _fetch_metric_data(  # pylint: disable=too-many-arguments
```
cwasicki: Why is it made private? This can be used downstream to fetch arbitrary metrics. The docstring was wrong though.

src/frequenz/data/microgrid/component_data.py
```diff
@@ -157,32 +169,121 @@ async def metric_data(  # pylint: disable=too-many-arguments
...
+            "chp",
+            "consumption",
+        ),
+        resampling_period: timedelta = timedelta(seconds=10),
+        metrics: dict[str, str] | None = None,
+        keep_components: bool = False,
+        splits: bool = False,
+        data_fetch_mode: Literal["power", "energy"] = "power",
+    ) -> pd.DataFrame | None:
+        """Return power data for component types of a microgrid.
```
cwasicki: If it's returning ac active power it should go into the `ac_active_power` function.

src/frequenz/data/microgrid/component_data.py
```diff
@@ -157,32 +169,121 @@ async def metric_data(  # pylint: disable=too-many-arguments
...
+                microgrid_id=microgrid_id,
+                start=start,
+                end=end,
+                component_types=component_types,
+                resampling_period=resampling_period,
+                metric=fetch_metrics["power"],
+                keep_components=keep_components,
+                splits=splits,
+            )
+
```
cwasicki, line 286: Can't we just use the `ac_active_energy_net` method instead of the code from here onward?

src/frequenz/data/microgrid/component_data.py
```diff
@@ -240,3 +334,118 @@ async def soc(  # pylint: disable=too-many-arguments
...
+        if delivered is None:
+            return consumed
+        return consumed.sub(delivered, fill_value=0)
+
+    @staticmethod
+    def _convert_cumulative_energy_to_power(
+        df: pd.DataFrame, *, resampling_period: timedelta
+    ) -> pd.DataFrame:
+        """Convert cumulative energy meter readings into power values."""
+        return df.diff().shift(-1) * (timedelta(hours=1) / resampling_period)
```
cwasicki: This might need more care in case the cumulative energy contains zeroes (e.g. on meter restarts).

PR #319, opened 2026-06-24
Add Battery Usecase plot

Mohammad-Tayyab-Frequenz:
This PR adds a dedicated reporting battery-usecase plot together with a standardized `create_battery_usecase_df()` helper to prepare the required input data. It also updates the battery usecase workflow to use German reporting labels, support optional PV area fills, and improve battery charge/discharge fill rendering and shared color handling.

This is how the battery usecase looks currently:

newplot (9)

### DIFF
--- a/src/frequenz/lib/notebooks/reporting/data_processing.py
+++ b/src/frequenz/lib/notebooks/reporting/data_processing.py
@@ -11,9 +11,11 @@ metrics used downstream for dashboards.
 Functions:
 ------------
 - Energy Report DataFrame Construction
-  - :func:`create_energy_report_df`: Builds a normalized energy report table with
+  - `create_energy_report_df`: Builds a normalized energy report table with
     unified naming, timezone conversion, grid import calculation, and
     component renaming based on a MicrogridConfig.
+  - `create_battery_usecase_df`: Builds the standardized battery-usecase table
+    used by the reporting plot helpers.
 
 Usage:
 -----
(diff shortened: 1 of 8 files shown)


cwasicki:
Just looking at the plot, if this is for Peak Shaving, why is PV relevant here at all? And prices? I would focus only on what is necessary. Or is the intention to have a general battery plot?

Maybe @cyiallou has some input too.

Mohammad-Tayyab-Frequenz:
@cwasicki I would change the name to battery usecase plot since this would be used to showcase all the usecases covered in that MID and not specifically just peak shaving.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -14,6 +14,8 @@
   - :func:`create_energy_report_df`: Builds a normalized energy report table with
     unified naming, timezone conversion, grid import calculation, and
     component renaming based on a MicrogridConfig.
+  - :func:`create_battery_usecase_df`: Builds the standardized battery-usecase table
```
cwasicki: `:func:` This isn't rendered in our docs, see https://frequenz-floss.github.io/frequenz-lib-notebooks/latest/reference/frequenz/lib/notebooks/reporting/data_processing/

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -36,6 +38,83 @@
...
+
+    Selects the battery-usecase input columns from the source DataFrame, renames
+    them to the standardized reporting schema, derives grid consumption without
+    battery support, computes reference peak lines, and splits battery power
+    flow into charging and discharging series.
+
+    Args:
+        energy_report_df: Reporting DataFrame containing the source columns for
+            timestamp, grid consumption, and battery power flow.
+        timestamp_col: Column name in ``energy_report_df`` containing the
```
cwasicki: We only use single ticks in our docs, double ticks are not needed.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -36,6 +38,83 @@
...
+            timestamps.
+        grid_consumption_col: Column name in ``energy_report_df`` containing
+            grid consumption with battery support.
+        battery_power_flow_col: Column name in ``energy_report_df`` containing
```
cwasicki: What is battery power flow? Is this just battery power?

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -36,6 +38,83 @@
...
+            battery power flow.
+        pv_power_col: Optional column name in ``energy_report_df`` containing
```
cwasicki: I guess the `power` is implied and not needed. Otherwise should be consistent for all names.

src/frequenz/lib/notebooks/reporting/data_processing.py
```diff
@@ -36,6 +38,83 @@
...
+    if pv_power_col and pv_power_col in energy_report_df.columns:
+        selected_cols.append(pv_power_col)
+        rename_map[pv_power_col] = "PV"
+
+    battery_usecase_df = energy_report_df[selected_cols].rename(
+        columns={
+            **rename_map,
+        }
+    )
+    battery_usecase_df["Netzbezug ohne Batterie"] = (
```
cwasicki: We shouldn't use German names in our code, only for visualization. I would address this with a mapping.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Removed :func: from the docstrings

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Updated.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Updated

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: Removed the power prefix in the name.

src/frequenz/lib/notebooks/reporting/data_processing.py
Mohammad-Tayyab-Frequenz: It is same as the battery col which contains the power flow through the battery (charging and discharging).

Should I update this as well to `battery_col`?

cwasicki: IMO yes.

Mohammad-Tayyab-Frequenz: Done

cwasicki (approved):

PR #334, opened 2026-07-07
Add day ahead prices

Mohammad-Tayyab-Frequenz:
This PR adds ENTSO-E day-ahead price support via a new dayahead.py helper, including API-key validation, timezone/window checks, and CSV output for fetched prices.

### DIFF
--- a/src/frequenz/lib/notebooks/dayahead.py
+++ b/src/frequenz/lib/notebooks/dayahead.py
@@ -1,0 +1,98 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Functions for CLI tool to interact with the day-ahead pricing API."""
+
+from __future__ import annotations
+
+import os
+from datetime import datetime
+
+import pandas as pd
+from entsoe import EntsoePandasClient  # type: ignore[attr-defined]
+
+
+def _normalize_prices(
+    prices: pd.Series | pd.DataFrame,
+) -> pd.Series:
+    """Normalize ENTSO-E day-ahead prices into a single named series.
+
+    Args:
+        prices: The raw response from `query_day_ahead_prices()`.
+
+    Returns:
+        A series named ``price`` with index name ``timestamp``.
+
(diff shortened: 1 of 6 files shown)


src/frequenz/lib/notebooks/dayahead.py
```diff
@@ -0,0 +1,123 @@
...
+            f"single-column DataFrame, got {type(prices).__name__}."
+        )
+
+    prices = prices.copy()
+    prices.name = "price"
+    prices.index.name = "timestamp"
+    return prices
+
+
+def list_day_ahead_prices(
```
cwasicki: Do we really need a function that just prints? It's the same as fetching with a single print statement

src/frequenz/lib/notebooks/dayahead.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+    start_ts = pd.Timestamp(start)
+    end_ts = pd.Timestamp(end)
+
+    if start_ts.tzinfo is None or end_ts.tzinfo is None:
+        raise ValueError("Please use timezone-aware datetimes for start and end.")
+
+    client = EntsoePandasClient(api_key=api_key)
+    return _normalize_prices(
```
cwasicki, line 96: Why is the normalization needed, i.e. when is the return value a data frame?

cwasicki (dismissed):

src/frequenz/lib/notebooks/dayahead.py
Mohammad-Tayyab-Frequenz: I picked it up from trading repo, I don't think we need it though.

src/frequenz/lib/notebooks/dayahead.py
Mohammad-Tayyab-Frequenz: We get a series from `query_day_ahead_prices` rather than a dataframe. The series is without a name so we have to add the column names to it.

cwasicki (approved):

